PR #97, opened 2025-08-22
Implement dynamic seeker and refactor update logic

nopeless:
https://github.com/user-attachments/assets/8c92efaa-3f07-42ef-82b8-2a34a12ec6a4

Checks if provider gave timespan of 0 when for MaxSeekTime to heuristically determine whether to show slider for song

The implementation is not really organized, but I am afraid to introduce new state tracking methods or make significant signature changes to existing functions.

Hence, I gave a minimal functionality implementation for now

Please read the code and let me know how you want this feature to be implemented (or just merge it if you think the code is fine)

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -336,7 +336,7 @@ public partial class MainWindow : MicaWindow
         {
             Dispatcher.Invoke(() =>
             {
-                if (!IsActive) return;
+                if (!IsActive || _isDragging) return;
 
                 UpdateSeekbarCurrentDuration(session.ControlSession.GetTimelineProperties().Position);
                 HandlePlayBackState(session.ControlSession.GetPlaybackInfo().PlaybackStatus);
@@ -574,7 +574,6 @@ public partial class MainWindow : MicaWindow
                     {
                         Seekbar.Maximum = timeline.MaxSeekTime.TotalSeconds;
                         SeekbarMaxDuration.Text = timeline.MaxSeekTime.ToString(timeline.MaxSeekTime.Hours > 0 ? @"hh\:mm\:ss" : @"mm\:ss");
-                        UpdateSeekbarCurrentDuration(timeline.Position);
                     }
                 }
             }
@@ -763,6 +762,7 @@ public partial class MainWindow : MicaWindow
 
     private void SeekbarUpdateUi(object? sender)
     {
+        return;
         if (!_seekBarEnabled || Visibility != Visibility.Visible || _isDragging) return;
         if (mediaManager.GetFocusedSession() is not { } session) return;
 
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -574,7 +574,7 @@ public partial class MainWindow : MicaWindow
                     {
                         Seekbar.Maximum = timeline.MaxSeekTime.TotalSeconds;
                         SeekbarMaxDuration.Text = timeline.MaxSeekTime.ToString(timeline.MaxSeekTime.Hours > 0 ? @"hh\:mm\:ss" : @"mm\:ss");
-                        SeekbarUpdateUi(null);
+                        UpdateSeekbarCurrentDuration(timeline.Position);
                     }
                 }
             }
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -562,7 +562,7 @@ public partial class MainWindow : MicaWindow
                     var timeline = mediaSession.ControlSession.GetTimelineProperties();
 
                     // State tracking
-                    bool mediaSessionSupportsSeekbar = timeline.MaxSeekTime.TotalSeconds >= 1.0; // Heruistics
+                    bool mediaSessionSupportsSeekbar = timeline.MaxSeekTime.TotalSeconds >= 1.0; // Heuristics
 
                     if (_mediaSessionSupportsSeekbar != mediaSessionSupportsSeekbar)
                     {
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -46,11 +46,12 @@ public partial class MainWindow : MicaWindow
     private bool _playerInfoEnabled = SettingsManager.Current.PlayerInfoEnabled;
     private bool _centerTitleArtist = SettingsManager.Current.CenterTitleArtist;
     private bool _seekBarEnabled = SettingsManager.Current.SeekbarEnabled;
-    
+    private bool _mediaSessionSupportsSeekbar = false;
+
     static Mutex singleton = new Mutex(true, "FluentFlyout"); // to prevent multiple instances of the app
     private NextUpWindow? nextUpWindow = null; // to prevent multiple instances of NextUpWindow
     private string currentTitle = ""; // to prevent NextUpWindow from showing the same song
-    
+
     private readonly int _seekbarUpdateInterval = 300;
     private readonly Timer _positionTimer;
     private bool _isActive;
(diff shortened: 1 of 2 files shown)


nopeless:
it appears that formatting got included in the commit as well, lmk if u want those gone too

nopeless:
[5cabe6d](https://github.com/unchihugo/FluentFlyout/pull/97/commits/5cabe6d177dee5fb72724709703fe21af8b42db6)
fixes a bug where if the media is paused seeker doesn't show correctly

nopeless:
so, I have been working on a refactor that removes the position timer completely because that was a lot more cleaner in terms of number of updates and had much less "glitchiness", but I'm not sure what things to test

unchihugo:
Hi, thank you for contributing!

This is a great addition to the current seek bar, I like the idea! By the way, is the forced refresh of the UI strictly necessary? 

Since you've labeled the PR as WIP, I won't merge it until you're ready to submit :)

> so, I have been working on a refactor ... but I'm not sure what things to test

We haven't set up any testing so far, just make sure it's free of bugs (or set up tests)!

nopeless:
@unchihugo thanks for giving the attention. I'll try to explain many types of issues that are present right now

When seek bar disappears, there exists extra empty space that doesn't look good. I considered two alternatives

A. Remove seekbar and show new flyout (implemented)
B. Keep seekbar but as empty bar

I chose A, and I'm not sure how you want it to be implemented. The important part is that it is totally possible to improve UX by detecting `MaxSeekTime`. I suggest you check my branch out.

If you have a particular vision or implementation in mind, please describe it, I'm willing to implement it myself

---

Additionally, I also tinkered with removing the arbitrary update cycle completely. I'll push a temporary commit at that point and let you test out the scrubbing and behavior.

FluentFlyoutWPF/MainWindow.xaml.cs
```diff
@@ -763,6 +762,7 @@ private void Seekbar_OnValueChanged(object sender, RoutedPropertyChangedEventArg
 
     private void SeekbarUpdateUi(object? sender)
     {
+        return;
```
nopeless: effectively disables this function entirely

nopeless:
I also have another feature I have been tinkering with that pauses all other media when a media plays (e.g. stopping a yt video when you start playing music). Implementation looks like this. I'd like to see this feature come by to the app, but I'm not ready to put out a new tracking issue or PR yet :3

```diff
index 593a8c8..e7fdb2e 100644
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -299,10 +299,15 @@ public partial class MainWindow : MicaWindow
         });
     }

-    private void CurrentSession_OnPlaybackStateChanged(MediaSession mediaSession, GlobalSystemMediaTransportControlsSessionPlaybackInfo? playbackInfo = null)
+    private async void CurrentSession_OnPlaybackStateChanged(MediaSession mediaSession, GlobalSystemMediaTransportControlsSessionPlaybackInfo? playbackInfo = null)
     {
         UpdateUI(mediaManager.GetFocusedSession());
         HandlePlayBackState(mediaManager.GetFocusedSession().ControlSession.GetPlaybackInfo().PlaybackStatus);
+
+        if (mediaSession.ControlSession.GetPlaybackInfo().PlaybackStatus == GlobalSystemMediaTransportControlsSessionPlaybackStatus.Playing)
+        {
+            await PauseOtherSessionsAsync(mediaSession);
+        }
     }

     private void MediaManager_OnAnyMediaPropertyChanged(MediaSession mediaSession, GlobalSystemMediaTransportControlsSessionMediaProperties mediaProperties)
@@ -895,4 +900,20 @@ public partial class MainWindow : MicaWindow
         }
         else if (SettingsManager.Current.nIconLeftClick == 1) ShowMediaFlyout();
     }
+
+    private async Task PauseOtherSessionsAsync(MediaSession playingSession)
+    {
+        foreach (var session in mediaManager.CurrentMediaSessions)
+        {
+            if (session.Key != playingSession.Id)
+            {
+                var playbackInfo = session.Value.ControlSession.GetPlaybackInfo();
+                if (playbackInfo.PlaybackStatus == GlobalSystemMediaTransportControlsSessionPlaybackStatus.Playing &&
+                    playbackInfo.Controls.IsPauseEnabled)
+                {
+                    await session.Value.ControlSession.TryPauseAsync();
+                }
+            }
+        }
+    }
 }
```

unchihugo:
Thank you for explaining the issue and your implementation to solve it!

Currently, I see that you've done it so that it'll need an extra refly since the `_mediaSessionSupportsSeekbar` check is done when the UI is updated, but maybe we could update the `_mediaSessionSupportsSeekbar` variable earlier so we're able to adjust the code in the `UpdateUILayout()` function first try?

I'm not sure how much more lightweight preventing a refly would be, but what do you think?

> I also have another feature I have been tinkering with that pauses all other media when a media plays

Sounds like a great feature! so the media player that's focused will be prioritized and pause all other sessions? btw, thank you for showing the wip implementation — I've never thought about using a foreach to loop through mediaManager.CurrentMediaSessions!

nopeless:
@unchihugo sorry i was flying back to the US. 

I am going to work on this in the weekends but first of all, is there a "testing program" that I can use for testing SMTC things? It is a bit hard to check behavior I can't seem to find any other applications that dont have the scrubber capability (other than the original musicbee dll). I just want another program to test consistent behavior

I'll check what I can do about the "double update" call you brought attention to.

You haven't really mentioned anything about disabling the timer, so I'd like a comment for that

I am not really a UI developer so if you add a toggle option for "pause other media when one plays" I can hook it up with the code (or you can implement that yourself too if you like the wip code I put in. I just think it can be cleaner)

unchihugo:
No worries, that's okay! I really appreciate your help anyway :)

> You haven't really mentioned anything about disabling the timer, so I'd like a comment for that

Could you clarify what you mean with disabling the timer? Do you mean the seek bar?

Edit: we can use another PR to implement the pausing feature, and I can commit an update to the UI in there!

nopeless:
@unchihugo https://github.com/unchihugo/FluentFlyout/pull/97#pullrequestreview-3149320626

FluentFlyoutWPF/MainWindow.xaml.cs
unchihugo: this function should stay enabled so that it constantly updates the seekbar's timer and position when the flyout is showing!

unchihugo:
hey, I did some testing and it seems like some media players that don't support seek bar will still have a `timeline.MaxSeekTime` near but not zero, such as `-00:00:00.0000010`. it should be an easy fix though!

you can live debug by adding these lines after 562 (after initializing `mediaSessionSupportsSeekbar`):
```c#
Debug.WriteLine("Media session supports seekbar: " + mediaSessionSupportsSeekbar);
Debug.WriteLine("Timeline max seek time: " + timeline.MaxSeekTime);
```

nopeless:
I'll add a check for `<1sec`, which should cover very short audio and negative values too

FluentFlyoutWPF/MainWindow.xaml.cs
nopeless: Ok, it turns out that some media applications I use automatically trigger `MediaManager_OnAnyTimelinePropertyChanged` every second for me... I think maybe a debounce is appropriate, as in if `MediaManager_OnAnyTimelinePropertyChanged` is fired, updates are suppressed for at least a second and a half, so that if the application manually dispatches updates every second, the timer code will never trigger leading to only a single update call site. Does this seem like a good implementation?

unchihugo:
Back from my trip, I've taken a look and it seems good now! Should we try merging this PR now?

nopeless:
@unchihugo I haven't tested extensively with chrome unfortunately, I only tested with firefox and one media player. 

I later realized this but Firefox and my mediaplayer actually sends timestamp updates every second, which I later accounted for in e06eeef but __you can go ahead and merge this after testing it in chrome__. It would be nice if there was testing guidelines somewhere

nopeless:
@unchihugo branch is ready

Issue #91, opened 2025-07-05
[BUG] Insert/Overwrite flyout shows the wrong state


xsm2:
**Describe the bug**
If you are in Insert mode when Fluent Flyout it's started the Insert/Overwrite flyout shows you the opposite state of the key instead of the correct one, when you are in Insert mode and press the Insert key the flyout shown is the "Insert mode", even if the current mode is Overflow.

**To Reproduce**
1. Close Fluent Flyouts
2. Change to Insert mode
3. Open Fluent Flyout
5. Open a text editor that supports the overwrite functionality like notepad or notepad++
6. Press the insert key
7. Write some text to check if the actual mode is the one the flyout said

**Expected behavior**
As with the rest of the lock keys, the Insert flyout should show the state after you have changed to, and not the one you have changed from, independently of the state that was active when the app was initialized. When changing from Insert to Overwrite the flyout should be "Overwrite mode" and when changing from Overwrite to Insert the flyout should be "Insert mode"

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version v1.4.1

PR #103, opened 2025-09-23
Add Insert key popup toggle and update insert flyout behavior

### LINKED ISSUE #71: [FEATURE REQUEST] Option to disable "Insert/Overwrite mode" flyout when using the Lock Keys Flyout
**Purpose:**  
Enabling Lock Keys Flyout also enables a flyout whenever the <kbd>INSERT</kbd> key is pressed. Using some shortcuts in terminal applications such as <kbd>CTRL</kbd>+<kbd>INSERT</kbd> (copy) / <kbd>SHIFT</kbd>+<kbd>INSERT</kbd> (paste), or using the key within some games (e.g. BeamNG.drive) causes a flyout to appear toggling between "Insert" and "Overwrite" mode, even when these modes are not relevant to the current application.

**Description:**  
An option to disable the "Insert Lock Key flyout" while also keeping other lock keys flyouts (Num, Caps, Scroll) enabled. 

Possibly a whitelist/blacklist setting per application, which would be interesting, even though it doesn't address the global state problem shown below.

**Additional context:**  
Since the <kbd>INSERT</kbd> key is technically not a "lock key", it doesn't have a global state. This causes its state to get desynced in certain situations.

Image
Image

unchihugo:
**New feature: Insert key popup control**

* Added a `LockKeysInsertEnabled` property to the `UserSettings` class, allowing users to enable or disable the Insert key popup.
* Updated the settings UI (`SettingsWindow.xaml`) to include a toggle switch for the new Insert key popup option, with explanatory text.
* Implemented the toggle switch logic in `SettingsWindow.xaml.cs` to update the setting when clicked.
* Modified the Insert key handler in `MainWindow.xaml.cs` to only show the flyout if `LockKeysInsertEnabled` is true.

**UI and text adjustments**

* Changed the Insert key flyout message in `LockWindow.xaml.cs` to always say "Insert pressed" instead of attempting to distinguish insert/overwrite mode.
* Updated the Save button text in the settings window to "Close" for clarity.

Closes #71 and #91

### DIFF
--- a/FluentFlyoutWPF/Classes/Settings/UserSettings.cs
+++ b/FluentFlyoutWPF/Classes/Settings/UserSettings.cs
@@ -22,8 +22,9 @@ public class UserSettings
     public bool nIconSymbol { get; set; }
     public bool DisableIfFullscreen { get; set; }
     public bool LockKeysBoldUI { get; set; }
-    public string LastKnownVersion { get; set; }
+    public string LastKnownVersion { get; set; } // for determining if user had updated to a new version
     public bool SeekbarEnabled { get; set; }
+    public bool LockKeysInsertEnabled { get; set; } // whether pressing insert key should display the LockKeys flyout
 
 
     public UserSettings()
@@ -50,5 +51,6 @@ public class UserSettings
         LockKeysBoldUI = true;
         LastKnownVersion = "";
         SeekbarEnabled = false;
+        LockKeysInsertEnabled = true;
     }
 }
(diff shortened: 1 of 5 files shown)


ANK-dev:
Is there an ETA for the next stable release?

unchihugo:
> Is there an ETA for the next stable release?

Thanks for reminding me - I've been busy, but I'll try to get it out ASAP, along with the other new features.

PR #224, opened 2025-12-20
Fixed thumbnail loading if they are loaded later.

### LINKED ISSUE #182: [BUG] Media flyout shows Chrome instead of YouTube Music for YouTube Music PWA
When playing YouTube Music installed as a PWA on Windows, the FluentFlyout media flyout shows "Google Chrome" instead of "YouTube Music" as the app name.

Note:
YouTube Music PWA runs within the browser engine, so the underlying process is chrome.exe. This may require special PWA detection logic. I'm not sure if this can be resolved, but I wanted to report it in case special handling is possible.

OS: Windows 11 Home
OS Build: 26200.7171
Windows Feature Experience Pack: 1000.26100.265.0
FluentFlyout Version: V2.3.0

DmitriySalnikov:
closes #182

If the player (in my case, the `Yandex Music` or `Youtube Music` web applications) did not cache the thumbnail image, then the standard browser icon is displayed instead. In fact, Chromium repeats the `OnAnyMediaPropertyChanged` event call several times (1-4 for me), but only the thumbnail changes in it, so due to current checks, the update does not occur.

I've added a few additional conditions to update the interface if the icon has been changed. This will cause extra window updates to happen, but it usually happened from 1 to 4 times for me and this should not significantly harm performance.

With this patch, `NextUpWindow` will be instantly replaced with the new one if the previous one has not closed yet, but the `Title` has already changed.

Also, all windows will update the interface if the thumbnail is updated while these windows are being displayed.

#### Old:

https://github.com/user-attachments/assets/82f25d5f-2071-4224-b433-854f1500d10f

#### New:

https://github.com/user-attachments/assets/8f4d3141-d41d-4e67-a3f8-18bfeaed0435

<sub>(here, the icon failed to load once, but this was most likely due to high load on the UI thread caused by recording or it only breaks like that when switching to the previous track. )</sub>

In fact, the old behavior didn't work so badly, but it very often displayed the standard icon (usually if the track has not been played yet). The same behavior can be achieved if you open the console on the player's page (<kbd>F12</kbd> or <kbd>Shift + Ctrl + I</kbd>) and turn off the cache.

browser_pbVoU6WFxP

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -28,7 +28,7 @@ namespace FluentFlyoutWPF;
 public partial class MainWindow : MicaWindow
 {
     private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
-    
+
     [DllImport("user32.dll")]
     public static extern void keybd_event(byte virtualKey, byte scanCode, uint flags, IntPtr extraInfo);
 
@@ -420,6 +420,7 @@ public partial class MainWindow : MicaWindow
 
     // for determining whether MediaPropertyChanged has no changes
     private string previousMediaProperty = "";
+    private string previousMediaPropertyThumbnail = "";
     private void MediaManager_OnAnyMediaPropertyChanged(MediaSession mediaSession, GlobalSystemMediaTransportControlsSessionMediaProperties mediaProperties)
     {
 #if DEBUG
@@ -434,11 +435,18 @@ public partial class MainWindow : MicaWindow
         var songInfo = mediaSession.ControlSession.TryGetMediaPropertiesAsync().GetAwaiter().GetResult();
         var playbackInfo = mediaSession.ControlSession.GetPlaybackInfo();
 
-        string check = songInfo.Title + songInfo.Artist + playbackInfo.PlaybackStatus + songInfo.Thumbnail;
+        string check = songInfo.Title + songInfo.Artist + playbackInfo.PlaybackStatus;
+        string checkThumbnail = songInfo.Thumbnail?.GetHashCode().ToString() ?? "Null";
+        bool onlyThumbnailChanged = false;
         if (previousMediaProperty == check)
+        {
+            onlyThumbnailChanged = true;
+            if (previousMediaPropertyThumbnail == checkThumbnail)
             return; // prevent multiple calls for the same song info
+        }
 
         previousMediaProperty = check;
+        previousMediaPropertyThumbnail = checkThumbnail;
 
         taskbarWindow?.UpdateUi(songInfo.Title, songInfo.Artist, Helper.GetThumbnail(songInfo.Thumbnail), playbackInfo.PlaybackStatus, playbackInfo.Controls);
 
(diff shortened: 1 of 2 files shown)


DmitriySalnikov:
> Just curious, what made you choose `songInfo.Thumbnail?.GetHashCode().ToString()` over `songInfo.Thumbnail`

Because when adding `songInfo.Thumbnail` to a string, only the class name was added to the string, something like `...IRandomAccessStreamReference`. Therefore, I take `GetHashCode()`, which differs for new images.

image

> and why did you move the thumbnail updating code in a new UpdateThumbnail() function inside NextUpWindow?

It is used in `MainWindow.MediaManager_OnAnyMediaPropertyChanged`.
At first, I thought about moving most of the code into a separate `UpdateUi` method, but then I decided it wasn't necessary because only the thumbnail can be updated. Nothing else is updated except for the thumbnail.
If other information changes, `NextUpWindow` is completely recreated.

unchihugo:
Ah okay, makes sense, thanks!

DmitriySalnikov:
@unchihugo I was just trying to finish the screen selection in my [display_select](https://github.com/DmitriySalnikov/FluentFlyout/tree/display_select) branch, but I ran into a very strange bug when the monitors have different DPI.

(from left 100% to the right 125%)

https://github.com/user-attachments/assets/827cda28-22b5-4580-9bbf-63c3e7e8f9a4

I'm not really sure what causes this behavior, but I'll try to fix it tomorrow. This branch should solve a couple of problems when people didn't have a window on the screen.

unchihugo:
> @unchihugo I was just trying to finish the screen selection in my [display_select](https://github.com/DmitriySalnikov/FluentFlyout/tree/display_select) branch, but I ran into a very strange bug when the monitors have different DPI.
> 
> I'm not really sure what causes this behavior, but I'll try to fix it tomorrow. This branch should solve a couple of problems when people didn't have a window on the screen.

That's certainly strange, and I have no idea why it would do that either. I've also noticed similar glitches when messing around with different displays, but I hadn't looked into it. By the way, are you using a Windows 10 taskbar? It's is styled differently.

PS: Thanks for investigating what might be one of the most popular requests for FluentFlyout :)

DmitriySalnikov:
> By the way, are you using a Windows 10 taskbar? It's is styled differently.

Overall, I am satisfied with Win11, but I don't like its various touch-oriented elements (with huge gaps and empty spaces). So I used Windhawk and applied several patches.

https://github.com/user-attachments/assets/0a373912-788f-4344-983a-a4dadae04afb

I have also moved the main Taskbar and notifications to the second monitor.
image

`FluentFlyout` provides functionality that I lost in Win11, but I miss the ability to move notifications to a second monitor.

---

With the help of `dnSpy`, I managed to find out that the `Window.Top` animation somehow causes constant DPI changes during window movement. Also, when `Top` changes, `Left` changes as well. And because of the DPI change, this causes the window to move along the `X` axis.
The `Window` class's private `_actualLeft` value moves from the set value of `-1904` to `-1.6` every time during animation. Perhaps this framework also does not understand how to work with multiple monitors 🤷‍♂️.

Issue #175, opened 2025-11-29
[BUG] Lock flyout text is cutted if too big


xsm2:
**Describe the bug**
I'm using FluentFlyout in Spanish, which usually is a language that needs more characters than English, and the Lock Flyout's text is not showed correctly.

**To Reproduce**
1. Change the language to Spanish
2. Enable the Lock Flyout
3. Press a lock key (ex: Shift)

**Expected behavior**
Either the text should be smaller or the widget bigger to accommodate the longer text

**Screenshots**

<img width="257" height="90" alt="Image" src="https://github.com/user-attachments/assets/f5d6ed39-32e1-455a-bb99-f8e0441f1f86" />

<img width="248" height="81" alt="Image" src="https://github.com/user-attachments/assets/b4a454f3-8bd7-4132-bc6e-9e9f2f0b08ec" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version v2.3.0

**Additional context**
I'm the person who has done the Weblate Spanish translation. Just the names that keys have printed on them on a keyboard are far longer in Spanish than in English, as they are already shortened: "Bloq Mayús" it's the short version of "Bloqueo Mayúsculas", "Bloq Despl" is "Bloqueo Desplazamiento" and "Bloq Num" is "Bloqueo Numérico"

unchihugo:
Hi, thank you for bringing this to my attention. I'll see what I can do about this.

Also, thank you for helping with the translations, I appreciate it!

PR #227, opened 2025-12-20
Variable Lock Flyout size for non English languages

xsm2:
This PR is related to the issue https://github.com/unchihugo/FluentFlyout/issues/175, which was fixed in  https://github.com/unchihugo/FluentFlyout/pull/188.

The fix that was made gave too much padding to the Spanish language and didn't account for other languages, the French text for example is even longer than the Spanish text and it's still wrong.

imagen

I have added a new size variable to the Localization manager that is recalculated each time a non English language is set to automatically change the width of the flyout. Which means there won't be any changes needed even if new languages are added.

imagen

### DIFF
--- a/FluentFlyoutWPF/Classes/LocalizationManager.cs
+++ b/FluentFlyoutWPF/Classes/LocalizationManager.cs
@@ -39,7 +39,7 @@ public static class LocalizationManager
     }
 
     public static LocalizationState Instance { get; } = new();
-    public static double maxLenght = 0;
+    public static double maxLength = 0;
 
     // current language code (first two letters) for easy access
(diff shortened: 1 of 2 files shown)


FluentFlyoutWPF/Classes/LocalizationManager.cs
```diff
@@ -39,6 +39,7 @@ protected void OnPropertyChanged([CallerMemberName] string? propertyName = null)
     }
 
     public static LocalizationState Instance { get; } = new();
+    public static double maxLenght = 0;
```
unchihugo: rename to maxLength

xsm2: Done!

Issue #200, opened 2025-12-08
[FEATURE REQUEST] Multi Monitor Pinning.


WobblySausage:
**Purpose:**  
Ability to position the Taskbar widget on a different screen. 
Media Flyout option to to allow it to appear on the monitor of our choosing. 

Additional option to pin the Media Flyout to always stay open on the monitor of our choosing. Serves as a great media controller for Spotify, I just wished that it didn't stay on one screen. Super annoying when gaming!

vinnha:
would loooove this

unchihugo:
I believe this is a duplicate of another issue, I'll close this one and link it to the original issue.

Issue #75, opened 2025-04-30
[BUG] The flyout won't appear when i play spotify


dwennbytes:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 11]
 - Version [e.g. v1.2.2]

**Additional context**
Add any other context about the problem here, such as Event Viewer logs.

Issue #30, opened 2024-11-09
[FEATURE REQUEST]


berkkrkc09:
**Purpose:**  
My notifications are displayed on my second monitor because I don't want any distractions on my main monitor. I was going to do the same with flyouts but it is not an option yet. This feature would be a great solution for intrusion just like the already built-in compact layout.

**Description:**  
I would love to be able to set the screen that the flyout will display on with individual toggles. Example: I would set media and next up flyout to display on 2nd monitor but the lock keys flyout would be on my 1st monitor.

**Additional context:**  
I use a Windhawk mod to move the notifications to my second monitor. It would be a great combo with this app if it had monitor selection.

agbrettpittman:
I found Fluent Flyout today and absolutely love it! Being able to choose a monitor would be super nice :)

WobblySausage:
Yeah I'd pay for the app if it had multi monitor selection, doesn't work for me in at the moment because gaming and full screen apps make it redundant unfortunately. 

To be able to add this to another screen would be perfect

WobblySausage:
Updated but can't see this new feature as shown in the mp4 files uploaded. I can see flyout position, widget position. Using 2.4.1

unchihugo:
> Updated but can't see this new feature as shown in the mp4 files uploaded. I can see flyout position, widget position. Using 2.4.1

It's not released as an update yet. If you don't want to wait, you can test out the latest updates by building it yourself!

unchihugo:
Just released v2.5.0, which gives you the ability to change which monitor the flyouts (and widget) displays on. Currently, the new setting applies to all three flyouts we have. We'll look into being able to manually choose a display for each flyout in the future.

WobblySausage:
Excellent. Purchased!

PR #231, opened 2025-12-21
Added the ability to select the monitor for display. `DPI Awareness` mode has been activated.

### LINKED ISSUE #155: [BUG]
description:
the media flyout don't appear anywhere, only the next up one work.

 - OS: Win11
 - Version 25H2

DmitriySalnikov:
fixes #155, probably #75 (if the user installed Spotify as a PWA)

Part of #30 (This PR changes the position of all windows), #200 (This PR does not affect the Widget)

* `DPI Awareness` mode has been activated.
* - All the application windows reported that they were using DPI Awareness `PerMonitorV2`, which was not working properly. Some windows displayed the contents correctly, but, for example, `NextUpWindow` blurred its contents. The solution was to add a `manifest` with the addition of the correct `dpiAwareness` tag.
* Added `workArea.Left` and `workArea.Top` to all layout options so that they are always located on the selected monitor.
* - Also, as a result, I had to convert all animation initialization to DPI-independent coordinates, which are multiplied by the monitor scale at the end.
* A new option has been added to the settings.
* Added the ability to get a list of available monitors.
* Fixed some window flickering.
* Fixed a few `NullReferenceException`s (not sure what caused them).

https://github.com/user-attachments/assets/d6f2af17-d68c-4279-81f9-d6f5dda695c9

https://github.com/user-attachments/assets/3020bd03-6e46-4520-b13f-b6971d55cc9d

---

<details>
<summary>About the taskbar</summary>

I also looked into how it would be possible to move `TaskbarWindow` to any of the taskbars, but to do that, I would need to somehow obtain a third taskbar (if possible)...
So maybe another time.
image

</details>

### DIFF
--- a/FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
+++ b/FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
@@ -15,7 +15,7 @@
     <System:String x:Key="CompactLayoutTitle">Compact Layout</System:String>
     <System:String x:Key="CompactLayoutDescription" xml:space="preserve">Slimmer and less intrusive layout
 hides repeat, shuffle and player info</System:String>
-    <System:String x:Key="SelectedMonitorTitle">Monitor for display</System:String>
+    <System:String x:Key="SelectedMonitorTitle">Flyout Target Display</System:String>
     <System:String x:Key="FlyoutPositionTitle">Flyout Position</System:String>
     <System:String x:Key="PositionBottomLeft" xml:space="preserve"> Bottom Left</System:String>
     <System:String x:Key="PositionBottomCenter" xml:space="preserve"> Bottom Center</System:String>
--- a/FluentFlyoutWPF/SettingsWindow.xaml
+++ b/FluentFlyoutWPF/SettingsWindow.xaml
@@ -100,15 +100,6 @@
                         <ui:CardControl Grid.Row="5" Icon="{ui:SymbolIcon ArrowMove24}" Margin="0,0,0,3">
                             <ui:CardControl.Header>
                                 <StackPanel Orientation="Vertical">
-                                    <TextBlock Text="{DynamicResource SelectedMonitorTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
-                                </StackPanel>
-                            </ui:CardControl.Header>
-                            <ComboBox x:Name="SelectedMonitorComboBox" SelectedIndex="{Binding SelectedMonitor, Mode=TwoWay}" >
-                            </ComboBox>
-                        </ui:CardControl>
-                        <ui:CardControl Grid.Row="6" Icon="{ui:SymbolIcon ArrowMove24}" Margin="0,0,0,3">
-                            <ui:CardControl.Header>
-                                <StackPanel Orientation="Vertical">
                                     <TextBlock Text="{DynamicResource FlyoutPositionTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
                                 </StackPanel>
                             </ui:CardControl.Header>
@@ -121,7 +112,7 @@
                                 <ComboBoxItem Content="{DynamicResource PositionTopRight}"/>
                             </ComboBox>
                         </ui:CardControl>
-                        <ui:CardControl Grid.Row="7" Icon="{ui:SymbolIcon Timer24}" Margin="0,0,0,12">
+                        <ui:CardControl Grid.Row="6" Icon="{ui:SymbolIcon Timer24}" Margin="0,0,0,12">
                             <ui:CardControl.Header>
                                 <StackPanel Orientation="Vertical">
                                     <TextBlock Text="{DynamicResource FlyoutStayDurationTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
@@ -135,7 +126,7 @@
                                 <TextBlock Text="{DynamicResource MillisecondsUnit}" FontSize="14" FontWeight="Regular" Margin="10,0,0,0" VerticalAlignment="Center"/>
                             </StackPanel>
                         </ui:CardControl>
-                        <ui:CardControl Grid.Row="8" Icon="{ui:SymbolIcon AlignLeft24}" Margin="0,0,0,3">
+                        <ui:CardControl Grid.Row="7" Icon="{ui:SymbolIcon AlignLeft24}" Margin="0,0,0,3">
                             <ui:CardControl.Header>
                                 <StackPanel Orientation="Vertical">
                                     <TextBlock Text="{DynamicResource CenterTitleArtistTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
@@ -144,7 +135,7 @@
                             </ui:CardControl.Header>
                             <controls:ToggleSwitch Name="CenterTitleArtistSwitch" IsChecked="{Binding CenterTitleArtist, Mode=TwoWay}"/>
                         </ui:CardControl>
-                        <ui:CardControl Grid.Row="9" Icon="{ui:SymbolIcon Subtitles24}" Margin="0,0,0,12">
+                        <ui:CardControl Grid.Row="8" Icon="{ui:SymbolIcon Subtitles24}" Margin="0,0,0,12">
                             <ui:CardControl.Header>
                                 <StackPanel Orientation="Vertical">
                                     <TextBlock Text="{DynamicResource ShowMediaPlayerNameTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
@@ -152,7 +143,7 @@
                             </ui:CardControl.Header>
                             <controls:ToggleSwitch Name="PlayerInfoSwitch" IsChecked="{Binding PlayerInfoEnabled, Mode=TwoWay}"/>
                         </ui:CardControl>
-                        <ui:CardControl Grid.Row="10" Icon="{ui:SymbolIcon ArrowRepeatAll24}" Margin="0,0,0,3">
+                        <ui:CardControl Grid.Row="9" Icon="{ui:SymbolIcon ArrowRepeatAll24}" Margin="0,0,0,3">
                             <ui:CardControl.Header>
                                 <StackPanel Orientation="Vertical">
                                     <TextBlock Text="{DynamicResource RepeatButtonTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
@@ -161,7 +152,7 @@
                             </ui:CardControl.Header>
                             <controls:ToggleSwitch Name="RepeatSwitch" IsChecked="{Binding RepeatEnabled, Mode=TwoWay}"/>
                         </ui:CardControl>
-                        <ui:CardControl Grid.Row="11" Icon="{ui:SymbolIcon ArrowShuffle24}" Margin="0,0,0,3">
+                        <ui:CardControl Grid.Row="10" Icon="{ui:SymbolIcon ArrowShuffle24}" Margin="0,0,0,3">
                             <ui:CardControl.Header>
                                 <StackPanel Orientation="Vertical">
                                     <TextBlock Text="{DynamicResource ShuffleButtonTitle}" FontSize="14" FontWeight="Regular" VerticalAlignment="Center"/>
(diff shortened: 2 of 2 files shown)


unchihugo:
Looks good so far! I'm glad you could fix the window position issue.

I can't test it tonight as I won't be on my computer, but looking at the code and your videos I have a few questions:

1. From the video, it looks like the Up Next window is still cut off for some reason, even though you've just made a PR to fix that (https://github.com/unchihugo/FluentFlyout/pull/223)
2. Will display 1 always be the main display (set in Windows)? Otherwise, we'll have to account for cases where the main screen != 1, with default settings setting it to 1.
3. Is the new SetVisibility method better than our current methods of setting topmost? If yes, should we adapt all windows to use this new function instead? (You don't have to bundle it in this PR)

Thanks for your time!

DmitriySalnikov:
> 1. From the video, it looks like the Up Next window is still cut off for some reason, even though you've just made a PR to fix that ([Consider the size of "Up next" when initializing `NextUpWindow`. #223](https://github.com/unchihugo/FluentFlyout/pull/223))

Oh, this video was recorded on the same day that I posted the other two PRs and it seems before I updated this branch.

https://github.com/user-attachments/assets/943554ed-f3cf-4007-a4ea-dee66abcbbc7

> 2\. Will display 1 always be the main display (set in Windows)? Otherwise, we'll have to account for cases where the main screen != 1, with default settings setting it to 1.

No, the primary monitor does not necessarily have to be the first one.
36fKYGLnuh
But I'm not sure how this affects the order of the screens after a reboot. In my OBS Notifier, one person had problems choosing a monitor, but only because he used `KVM switch`.
Even if I extend `MONITORINFO` to `MONITORINFOEX`, I can only get `\\\\.\\DISPLAY*`
image
And if we use even more structures and enumerators, we can obtain the following data:
y4l2iypX6C

However, I do not know which of these values are unchanged under all changing conditions. Probably none, so for now I would leave `EnumDisplayMonitors` and saving the index.

Resetting to the Primary monitor takes place here: https://github.com/unchihugo/FluentFlyout/pull/231/files#diff-fa57c105be0e96f1831f0a78a9a8007180b8992354f43085e3afef50f05c36cdR231-R232

> 3\. Is the new SetVisibility method better than our current methods of setting topmost? If yes, should we adapt all windows to use this new function instead? (You don't have to bundle it in this PR)

I'm not sure where else `Visibility.Hidden` is used, but in this case, using it resulted in several frames of delay until the window became invisible. That's why I started using `SetVisibility`, this method instantly hides the window, preventing it from being displayed in the corner of the screen. The most noticeable flickering occurred when displaying `NextUpWindow`, because it is recreated each time and did not retain the "hidden" state from the previous animation.

I also used `SetVisibility` [here](https://github.com/unchihugo/FluentFlyout/pull/231/files#diff-1dc2a643641d41f9375d1d0eaccc42727996fa317cc8a151618c7f81d682978aR522) because of the very strange stretching of the window before closing.

https://github.com/user-attachments/assets/695063e5-0758-4cab-afd4-92ef69112899

This may also be due to the different DPI on the monitors 🤷‍♂️

---

<details>
<summary>About device names</summary>

I'll leave this code for fetching device IDs here so I don't lose it if I need it again.

```diff
diff --git a/FluentFlyoutWPF/Classes/WindowHelper.cs b/FluentFlyoutWPF/Classes/WindowHelper.cs
index bece504..2cad97f 100644
--- a/FluentFlyoutWPF/Classes/WindowHelper.cs
+++ b/FluentFlyoutWPF/Classes/WindowHelper.cs
@@ -81,6 +81,45 @@ public static class WindowHelper
         }
     }
 
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode, Pack = 4)]
+    private struct MONITORINFOEX
+    {
+        public int cbSize;
+        public RECT rcMonitor;
+        public RECT rcWork;
+        public int dwFlags;
+
+        [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 32)]
+        public string? szDevice;
+
+        public static implicit operator MonitorInfo(MONITORINFOEX other)
+        {
+            var monitor = new MonitorInfo
+            {
+                monitorArea = other.rcMonitor,
+                workArea = other.rcWork,
+                isPrimary = (other.dwFlags & MONITORINFOF_PRIMARY) != 0
+            };
+
+            return monitor;
+        }
+    }
+
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode, Pack = 4)]
+    private struct DISPLAY_DEVICE
+    {
+        public int cbSize;
+        [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 32)]
+        public string? DeviceName;
+        [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 128)]
+        public string? DeviceString;
+        public int StateFlags;
+        [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 128)]
+        public string? DeviceID;
+        [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 128)]
+        public string? DeviceKey;
+    }
+
     public struct MonitorInfo
     {
         public Rect monitorArea;
@@ -95,8 +134,11 @@ public static class WindowHelper
     [DllImport("user32.dll")]
     private static extern bool EnumDisplayMonitors(IntPtr hdc, IntPtr lprcClip, MonitorEnumProc lpfnEnum, IntPtr dwData);
 
-    [DllImport("user32.dll", CharSet = CharSet.Auto)]
-    private static extern bool GetMonitorInfo(IntPtr hMonitor, ref MONITORINFO lpmi);
+    [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+    private static extern bool EnumDisplayDevices(string? lpDevice, uint iDevNum, ref DISPLAY_DEVICE lpDisplayDevice, uint dwFlags);
+
+    [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+    private static extern bool GetMonitorInfo(IntPtr hMonitor, ref MONITORINFOEX lpmi);
 
     [DllImport("user32.dll")]
     private static extern IntPtr MonitorFromWindow(IntPtr hwnd, int dwFlags);
@@ -156,7 +198,7 @@ public static class WindowHelper
 
     private static MonitorInfo getMonitorInfoInternal(IntPtr hMonitor)
     {
-        MONITORINFO info = new() { cbSize = Marshal.SizeOf<MONITORINFO>() };
+        MONITORINFOEX info = new() { cbSize = Marshal.SizeOf<MONITORINFOEX>() };
 
         if (GetMonitorInfo(hMonitor, ref info))
         {
@@ -179,6 +221,45 @@ public static class WindowHelper
         return new MonitorInfo();
     }
 
+    public static void GetDisplayDevices()
+    {
+        for (uint i = 0; ; i++)
+        {
+            DISPLAY_DEVICE dd = new DISPLAY_DEVICE() { 
+                cbSize = Marshal.SizeOf<DISPLAY_DEVICE>(),
+            };
+
+            if (EnumDisplayDevices(null, i, ref dd, 0 /*1 EDD_GET_DEVICE_INTERFACE_NAME*/))
+            {
+                const int DISPLAY_DEVICE_ACTIVE = 0x1;
+                const int DISPLAY_DEVICE_PRIMARY_DEVICE = 0x4;
+
+                bool isActive = (dd.StateFlags & DISPLAY_DEVICE_ACTIVE) != 0;
+                bool isPrimary = (dd.StateFlags & DISPLAY_DEVICE_PRIMARY_DEVICE) != 0;
+
+                if (dd.DeviceName != "\\")
+                {
+                    var a = isActive;
+                }
+
+                DISPLAY_DEVICE res = new DISPLAY_DEVICE()
+                {
+                    cbSize = Marshal.SizeOf<DISPLAY_DEVICE>(),
+                };
+
+                if (EnumDisplayDevices(dd.DeviceName, 0, ref res, 0 /*1 EDD_GET_DEVICE_INTERFACE_NAME*/))
+                {
+                    var b = dd.StateFlags;
+                }
+            }
+            else
+            {
+                break;
+            }
+        }
+
+    }
+
     public static MonitorInfo GetMonitor(Window window)
     {
         var handle = new WindowInteropHelper(window).Handle;
@@ -199,6 +280,7 @@ public static class WindowHelper
             },
             IntPtr.Zero
         );
+        GetDisplayDevices();
 
         return result;
     }

```

</details>

unchihugo:
Thanks for the detailed explanation, I appreciate it! The display switching works on my end as well.

One issue: using the build of this branch, I've noticed that I, as well, get taskbar hangs and crashes after right-clicking the taskbar. This is not observed on the origin/master branch. Something must've been introduced that makes it hang?

I did manage to get an error log, although I doubt that this is error is actually the cause — it seems like the result of something else that make taskbar hang that subsequently triggered this error.
```
ERROR|FluentFlyout.Windows.TaskbarWindow|COM error retrieving taskbar widgets button Rect.|System.Runtime.InteropServices.COMException (0x8000FFFF): Catastrophic failure (0x8000FFFF (E_UNEXPECTED))
   at MS.Internal.Automation.UiaCoreApi.CheckError(Int32 hr)
   at MS.Internal.Automation.UiaCoreApi.UiaFind(SafeNodeHandle hnode, UiaFindParams findParams, Condition findCondition, UiaCacheRequest request)
   at System.Windows.Automation.AutomationElement.Find(TreeScope scope, Condition condition, UiaCacheRequest request, Boolean findFirst, BackgroundWorker worker)
   at System.Windows.Automation.AutomationElement.FindFirst(TreeScope scope, Condition condition)
   at FluentFlyout.Windows.TaskbarWindow.GetTaskbarWidgetRect(IntPtr taskbarHandle) in C:\Users\hu8go\Documents\GitHub\FluentFlyout\FluentFlyoutWPF\Windows\TaskbarWindow.xaml.cs:line 679
   ```
   
EDIT: I've also noticed that switching DPI results in incorrectly scaled windows and the taskbar widget. This seems like a harder issue to tackle than I thought.

DmitriySalnikov:
> One issue: using the build of this branch, I've noticed that I, as well, get taskbar hangs and crashes after right-clicking the taskbar. This is not observed on the origin/master branch. Something must've been introduced that makes it hang?

This is strange, since I have both branches hanging the same way.

https://github.com/user-attachments/assets/56063064-38e4-4cd0-ad0a-9f975ca4a216

But I forgot to mention one strange freeze before I enabled `dpiAwareness`: when the second monitor displaying the widget was zoomed in significantly, the entire application would freeze before it even launched properly. Immediately after adding `dpiAwareness` the hang disappeared.

> EDIT: I've also noticed that switching DPI results in incorrectly scaled windows and the taskbar widget. This seems like a harder issue to tackle than I thought.

Can you try to completely clear the `FluentFlyoutWPF/bin` and `FluentFlyoutWPF/obj` folders?

Maybe this crash will disappear too

> ERROR|FluentFlyout.Windows.TaskbarWindow|COM error retrieving taskbar widgets button Rect.|System.Runtime.InteropServices.COMException (0x8000FFFF): Catastrophic failure (0x8000FFFF (E_UNEXPECTED))

unchihugo:
>>EDIT: I've also noticed that switching DPI results in incorrectly scaled windows and the taskbar widget. This seems like a harder issue to tackle than I thought.
>
>Can you try to completely clear the FluentFlyoutWPF/bin and FluentFlyoutWPF/obj folders?

I'm still observing these issues (below an example of switching from 100% to 150% DPI scaling), but I'm fine with leaving this problem in the program for now as it's not incredibly important. I've stopped getting taskbar hangs/crashes for now — it looks to be quite random. 

image

SetVisibility() is indeed more reliable than `this.Visiblity`. From my own testing, it solved an issue with the LockWindow acrylic on first flyout.

I'll test this build a bit more before merging so that we can be sure that nothing unexpected is introduced.

DmitriySalnikov:
I was able to launch a VM with two monitors, put the `GitHub Release` build on it, set one monitor to 150% and the main one to 125%.
image

If `FluentFlyout` was running when the scaling changed, it needs to be restarted because even Task Manager remains blurred without restarting.
But all `FluentFlyout` windows look sharp and open in the right places.

Screenshot (1)

> image


Yes, my windows also break in the same way if I don't restart `FluentFlyout`. I'm not sure if it's possible to fix this. Most likely, the problem lies in WPF itself. They have a lot of DPI-related issues in their repository.

unchihugo:
Just moved the settings around and rewrote the setting title for more clarity. I'll merge now!

unchihugo (approved):

unchihugo:
Actually, are you planning on having a display setting for every flyout? In that case, the new layout I just pushed might not make sense.

DmitriySalnikov:
I've been thinking about this. We can add a monitor selection to each element and not even indent the center position if `LockWindow` is on another screen. But right now, I'm thinking about fixing the `Bottom Right` position of the widget (broken if the main taskbar is moved to another monitor) and going back to my code.

If necessary, these changes can be reverted later.

unchihugo:
Alright, let me know if you want the branch to be merged. I'm updating the app now as I'd like to have PR https://github.com/unchihugo/FluentFlyout/pull/228 out as soon as possible, so this will most likely make it to the update after this one.

DmitriySalnikov:
I think you can merge it now. Anyway, to select monitors for other elements, we need the functionality to get a monitor from this PR.

Issue #195, opened 2025-12-06
[BUG] Taskbar widget not positioned correctly vertically (2-in-1 PCs)


JackVac:
**Describe the bug**
The taskbar media widget doesn't align correctly on in the taskbar verticaly

Edit: So my laptop is a 360 and when I flip it into tablet mode the widget seems to be aligned correctly. It's like the widge stays stuck into tablet mode even if I execute fluent flyout while I'm not into tablet mode.

**To Reproduce**
I have no idea how to reproduce this issue

**Screenshots**
https://imgur.com/a/Z8rIdeo

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version 25H2

unchihugo:
Hi, thank you for reporting! 

I have a few questions:
1. Has the display scale ever changed (for example manually or when you launch a game that changes your scaling) during the time you've had FluentFlyout on? 
2. Does the widget go back in place when you restart the app?

JackVac:
No the display scale never changed during the time of using FluentFlyout. I just installed FluentFlyout and at the begining it wouldn't even show in the taskbar, but after restarting my laptop it showed in the taskbar but as you can see it now shows in the taskbar but not propely. Also no restarting the app doesn't fix the problem

unchihugo:
Do you open FluentFlyout using Windows Search? If so, try opening FluentFlyout via the start icon, as some users have reported issues opening via Windows Search.

If not, do you have any other taskbar extension enabled (e.g. Windhawk, ExplorerPatcher)?

JackVac:
Opening it by another method doesn't work. Also, I have no other taskbar extensions.

redolencetech:
+1 for same issue

JackVac:
I tried it on my gaming PC yesterday and I don't have the same weird aligment issue I have here on my laptop.

JackVac:
So my laptop is a 360 and when I flip it into tablet mode the widget seems to be aligned correctly. It's like the widge stays stuck into tablet mode even if I execute fluent flyout while I'm not into tablet mode.

ElNuevoIRONMAN:
Same issue right here, and as JackVac says, I've just installed the app, and it doesn't work properly the widget

unchihugo:
@ElNuevoIRONMAN @redolencetech do you also use a special kind of monitor like a surface / tablet / rotatable display?

JackVac:
[...]

PR #238, opened 2025-12-23
Fix `TaskbarWindow` positioning.

DmitriySalnikov:
When moving the main Taskbar from the main monitor, the widget in the `Bottom right` position goes beyond the screen boundaries.
This PR corrects `physicalLeft` and `physicalTop` to use truly global monitor coordinates, but then converts them to Taskbar coordinates.

#### Old:

https://github.com/user-attachments/assets/d7008d93-e5f4-472a-bcbd-44f7a9c3a45d

#### New:

https://github.com/user-attachments/assets/4274046b-bef9-4e83-ae4a-0c08b570b67a

### DIFF
--- a/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
+++ b/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
@@ -57,6 +57,9 @@ public partial class TaskbarWindow : Window
     [DllImport("user32.dll", SetLastError = true)]
     private static extern int GetWindowLong(IntPtr hWnd, int nIndex);
 
+    [DllImport("user32.dll")]
+    private static extern uint GetDpiForWindow(IntPtr hMonitor);
+
     [DllImport("user32.dll", SetLastError = true)]
     private static extern bool SetWindowPos(IntPtr hWnd, IntPtr hWndInsertAfter, int X, int Y, int cx, int cy, uint uFlags);
 
@@ -102,8 +105,6 @@ public partial class TaskbarWindow : Window
     private string _cachedArtistText = string.Empty;
     private double _cachedTitleWidth = 0;
     private double _cachedArtistWidth = 0;
-    private double _dpiScaleX;
-    private double _dpiScaleY;
     private IntPtr _trayHandle;
     private AutomationElement? _widgetElement;
     // reference to main window for flyout functions
@@ -123,10 +124,6 @@ public partial class TaskbarWindow : Window
         DataContext = SettingsManager.Current;
 
         _hitTestTransparent = new SolidColorBrush(System.Windows.Media.Color.FromArgb(1, 0, 0, 0));
-
-        // initialize here in case we want to restart the window
-        _dpiScaleX = 0;
-        _dpiScaleY = 0;
 
         _timer = new DispatcherTimer();
         _timer.Interval = TimeSpan.FromMilliseconds(1500); // slow auto-update for display changes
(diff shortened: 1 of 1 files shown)


DmitriySalnikov:
Oh, switching monitors with different DPI can still cause some problems.

image
image

---

Fixed

unchihugo:
Hi, thanks for the PR! Glad you could figure out how to make it work. Just wondering, is there currently a way to test this out without using Windhawk?

Also, I had a lot of issues trying to figure out how to adapt the taskbar widget to DPI changes. I had no idea it was that simple!

DmitriySalnikov:
> Just wondering, is there currently a way to test this out without using Windhawk?

Actually, I don't know of any other ways at the moment. But it will be relevant if someone (I'm not sure if it will be me😵) adds the ability to select a monitor for the widget.

unchihugo:
> > Just wondering, is there currently a way to test this out without using Windhawk?
> 
> Actually, I don't know of any other ways at the moment. But it will be relevant if someone (I'm not sure if it will be me😵) adds the ability to select a monitor for the widget.

That's okay, I've actually found a way to inject the widget into the 2nd taskbar, although making it work with a 3rd monitor (and 4th, 5th, etc.) is harder than I thought. I've read that all taskbars other than the main one is a "2nd" taskbar.

DmitriySalnikov:
> That's okay, I've actually found a way to inject the widget into the 2nd taskbar, although making it work with a 3rd monitor (and 4th, 5th, etc.) is harder than I thought. I've read that all taskbars other than the main one is a "2nd" taskbar.

Yeah, I've already done some research 😅. And I figured out how to implement this, but it bothers me that I have to get a list of all windows every 1.5 seconds.
I'll probably try to implement it tomorrow.

---

Oh, and I forgot to ask, what to do with WinAPI, which are partially duplicated in the widget and WindowHelper class?

unchihugo:
> Yeah, I've already done some research 😅. And I figured out how to implement this, but it bothers me that I have to get a list of all windows every 1.5 seconds.

Hmm, why would you need to get a list of all windows every 1.5s?

> Oh, and I forgot to ask, what to do with WinAPI, which are partially duplicated in the widget and WindowHelper class?

We should probably consolidate them to the WindowHelper class

DmitriySalnikov:
> Hmm, why would you need to get a list of all windows every 1.5s?

Because currently `FindWindow("Shell_TrayWnd", null)` is called every 1.5 seconds. But I think it will be possible to cache the list of taskbars found, and I can try using window search in the same thread as the main taskbar.

unchihugo:
> > Hmm, why would you need to get a list of all windows every 1.5s?
> 
> Because currently `FindWindow("Shell_TrayWnd", null)` is called every 1.5 seconds. But I think it will be possible to cache the list of taskbars found, and I can try using window search in the same thread as the main taskbar.

The FindWindow call only takes around 30 nanoseconds (or 0.00000003 seconds) in debug mode so I don't think that optimization is strictly necessary, but feel free to cache it!

EDIT: by the way are you planning on adding the taskbar selector? Just asking so we don't work on the same thing :)

DmitriySalnikov:
> EDIT: by the way are you planning on adding the taskbar selector? Just asking so we don't work on the same thing :)

I think so, I'll try to implement the taskbar selection.
If I lose interest in this task tomorrow, I'll let you know.

DmitriySalnikov:
> The FindWindow call only takes around 30 nanoseconds (or 0.00000003 seconds) in debug mode so I don't think that optimization is strictly necessary, but feel free to cache it!

Well, searching for all additional taskbars among all windows when there is only the main taskbar takes significantly longer.
image

https://github.com/user-attachments/assets/e0bd8d1e-5812-4cd7-a73b-5a421bb74ffc

I didn't even think that creating a virtual monitor would be so easy.

unchihugo:
> Well, searching for all additional taskbars among all windows when there is only the main taskbar takes significantly longer.

That's true, through my testing it's ~0.3ms every 1.5s, which is still really efficient, right? Another option would be to listen to WM_DISPLAYCHANGE and refresh the list that way. Perhaps this functionality and the monitor data you've written a few days ago could be put in a new class as it could be useful for the latter as well.

DmitriySalnikov:
> Perhaps this functionality and the monitor data you've written a few days ago could be put in a new class as it could be useful for the latter as well.

I think I won't add new classes yet, because it will require changing a lot of lines. I'll continue to expand `TaskbarWindow` and `WindowHelper`, and then you decide for yourself what to put where and combine it, okay?

DmitriySalnikov:
This PR will most likely fix #195.

unchihugo (approved):

PR #239, opened 2025-12-23
Fixed crash when right-clicking on a widget.

### LINKED ISSUE #203: [BUG] The taskbar widget crashes the entire desktop when right-clicking, and no longer has style when hovering over it
**Describe the error**
As stated in the title.

**To reproduce it**
1. Right-click on the widget.
2. Once frozen, click the Windows button (taskbar or keyboard).
3. Then the entire desktop crashes.

**Expected behavior**
It should display the style it had on the first day when hovering and should not crash the desktop when right clicking it.

**Screenshots**

https://github.com/user-attachments/assets/3b5f4e25-c42d-43b1-8dfc-ae722032c235

**Desktop (complete the following information):**
- Operating system: Windows 11
- Version: v2.3.5

**Additional info**
I have disabled all my tweaks (Rainmeter, Windhawk, TranslucentTB, NexusUltimate, WallpaperEngine) and still happens.

DmitriySalnikov:
fixes #203

Fixes deadlock (or something else?) when right-clicking on the widget.

At first, I thought the problem was specifically with right-clicking, so I used a different method of processing messages, and that helped with the freezing caused by `Click on empty taskbar space` from `Windhawk`. Old behavior: https://github.com/unchihugo/FluentFlyout/issues/203#issuecomment-3680876153
But `Nilesoft Shell` continued to cause freezes.

The solution turned out to be extremely simple to implement! but very difficult to trace the cause.
It seems that both of these mods also use `UI Automation` for their work, which leads to freezing. `UI Automation` sends message `0x003D`, without which it cannot function. Therefore, we can just prevent this message from propagating!

#### New behavior:

https://github.com/user-attachments/assets/b9680898-d9cf-488b-8fc4-9887dfd36b33

### DIFF
--- a/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
+++ b/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
@@ -130,6 +130,27 @@ public partial class TaskbarWindow : Window
         Show();
     }
 
+    protected override void OnSourceInitialized(EventArgs e)
+    {
+        base.OnSourceInitialized(e);
+        HwndSource source = (HwndSource)PresentationSource.FromDependencyObject(this);
+        source.AddHook(WindowProc);
+    }
+
+    private static IntPtr WindowProc(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lParam, ref bool handled)
+    {
+        // Some interface mods may collect information from all windows associated with the taskbar,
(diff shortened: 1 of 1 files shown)


unchihugo (approved):
Nice find! I had no idea about this, but it's great that you've managed to fix the issue. I'll merge the PR, thank you!

PR #242, opened 2025-12-24
Added the ability to select the monitor on which the widget will be displayed.

DmitriySalnikov:
fixes #200 (it's already closed, but this PR adds a taskbar selection)

* Renamed the `SelectedMonitor` setting to `FlyoutSelectedMonitor`.
* - #231 hasn't been released yet, so I think it's safe.
* Added `TaskbarWidgetSelectedMonitor` setting and corresponding field in the settings window.
* Replaced `FindWindow("Shell_TrayWnd", null)` with the new function `GetSelectedTaskbarHandle()`.
* This PR also fixes the widget's pinning to the taskbar, which may not occur if Explorer has been reloading for a long time.

Demo: https://github.com/unchihugo/FluentFlyout/pull/238#issuecomment-3690105083

---

If you also need it, then I used [Virtual-Display-Driver](https://github.com/VirtualDrivers/Virtual-Display-Driver) to add virtual monitors.

PR #244, opened 2025-12-25
Fixed the selection of the main taskbar when the widget is right-aligned.

DmitriySalnikov:
Since the main taskbar may not be on monitor 0 (and monitor 0 is not always the primary monitor), we should compare the taskbar handle with the `Shell_TrayWnd` handle.

* If the tray is not found, the fallback will be on the right side of the taskbar.
* Added caching for `GetSystemTrayRect`. https://github.com/unchihugo/FluentFlyout/pull/242#issuecomment-3690982105

#### Before

https://github.com/user-attachments/assets/4c86adb8-e89b-4833-8c3e-aec7bc24809b

#### After

https://github.com/user-attachments/assets/5a91c2be-a103-4278-a53a-b0cc032a0f9d

### DIFF
--- a/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
+++ b/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
@@ -817,6 +817,60 @@ public partial class TaskbarWindow : Window
     //    }
     //}
 
+    private (bool, Rect) GetTaskbarXamlElementRect(IntPtr taskbarHandle, ref AutomationElement? elementCache, string elementName)
+    {
+        try
+        {
+            // reset if monitor changed
+            if (_lastSelectedMonitor != SettingsManager.Current.TaskbarWidgetSelectedMonitor)
+                elementCache = null;
+
+            // find widget in XAML
+            if (elementCache == null)
+            {
+                AutomationElement root = AutomationElement.FromHandle(taskbarHandle);
+
+                elementCache = root.FindFirst(TreeScope.Descendants,
+                    new PropertyCondition(AutomationElement.AutomationIdProperty, elementName));
+            }
+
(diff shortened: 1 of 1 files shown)


unchihugo:
Thank you! I had totally forgotten about the caching (also looking back at the code I realize I left some debug code in there). 

By the way, the caching implementation differs from how GetTaskbarWidgetRect() caches the widget, so it might be a bit confusing for first time contributors. Do you think it needs to change?

DmitriySalnikov:
> By the way, the caching implementation differs from how GetTaskbarWidgetRect() caches the widget, so it might be a bit confusing for first time contributors. Do you think it needs to change?

Are you talking about using `tuple`s vs `out`? I don't really care either way. In this case, the only difference is in code style. Using `out` seems a little more natural to me as a C/C++ programmer. And tuples are something more modern.

unchihugo:
In essence, GetTaskbarWidgetRect() checks for cache inside the function, but the new code checks for cache before GetSystemTrayRect() gets called.

DmitriySalnikov:
Oh, in that sense. Actually, if it is possible to cache `AutomationElement trayElement` and this allows you to get the actual bounds with a slight delay, that would be even better.

For me, requesting the actual boundaries (`trayElement.Current.BoundingRectangle`) takes about 0.3 ms, which is quite acceptable.

Will you do it yourself?

unchihugo:
Sure, I'll handle it. Thanks for your help :)

unchihugo:
@DmitriySalnikov Does the functionality still work fine on your end (when switching main taskbars)?

DmitriySalnikov:
> Does the functionality still work fine on your end (when switching main taskbars)?

Yes, everything looks right .

These two methods seem to me to be absolutely identical, and I think they should be combined.

image

PR #255, opened 2025-12-27
Add basic tooltip functionality to Taskbar and MainWindow

### LINKED ISSUE #248: [FEATURE REQUEST] Show a tooltip above the widget when the title or artist name is too long
**Purpose:**  
Add the option to enable a tooltip if the title or artist name does not fit in the widget window.

Image

**Description:**  
Sometimes the titles don't fit in the widget, and to see the entire title, I have to open the player window (Flyout usually cuts the title even more).

Image Image

https://github.com/unchihugo/FluentFlyout/issues/51#issuecomment-2701380725 😂

**Additional context:**  
The tooltip can be made in the Mica/Acrylic style (<sub>it seems impossible</sub>) and centered at the top of the widget rather than at the mouse position.

unchihugo:
Adds tooltip by hovering over title or artist name — displaying full title and artist — in the Taskbar Widget and Media Flyout windows.

Closes #248

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -177,7 +177,7 @@ public partial class MainWindow : MicaWindow
             LocalizationManager.ApplyLocalization();
             // show settings to new users
             string previousVersion = SettingsManager.Current.LastKnownVersion;
-            if (previousVersion == "")
+            if (previousVersion == string.Empty)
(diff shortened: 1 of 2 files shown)


DmitriySalnikov:
As Copilot already pointed out in the last comment, the tooltip may duplicate the text several times, for example, due to cover updates in the browser player.
imageimage
In the first image, the first track played was added to the text that appears when the player is opened.

PR #268, opened 2026-01-07
An attempt to fix infinite explorer crashing issue with dynamic taskbar creation check

mclt0568:
Instead of relying on timeout (1000ms) to wait for taskbar to become stable (which in many case stabilisation takes more than a second), now relies `MainWindow.ExplorerRestarting` as a flag to determine if taskbar is stable

The `ExplorerRestarting` flag is flipped to true when `WM_TASKBARCREATED` is received, then flipped back to false when `WaitForExplorerReadyAsync()` returns, indicating its ready.

`WaitForExplorerReadyAsync()` works by probing taskbar every 200ms, checking if window exists (FindWindow()) and has a valid geometry. This happens until taskbar is ready or timed out (defaulted to 8000ms).

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -1287,7 +1287,7 @@ public partial class MainWindow : MicaWindow
         }
     }
 
-    private async Task<bool> WaitForExplorerReadyAsync(int timeoutMs = 8000)
+    private async Task<bool> WaitForExplorerReadyAsync(int timeoutMs = 60000)
     {
         var sw = Stopwatch.StartNew();
 


mclt0568:
that was brutal 😿 
tho please actually double check my changes doesn't break the app in other aspects, since im not extremely familiar with win API while attempting the fix..

unchihugo:
Hi, thank you for contributing, I appreciate it!

Quickly looking through the code the implementation looks great, but I noticed that you left a lot of commented code in and removed some try catch blocks and logging. If possible, could you add them back in areas where needed?

Other than that, do you think we should keep ExplorerRestarting flag to true even if the 8 second loop has passed? 

Thanks again!

mclt0568:
I've updated the comments and added the try-catch block back.

I think if taskbar isnt ready after 8 seconds is kinda sus, either the taskbar is never ready, maybe due to other plugins that modifies the taskbar and causing issue?

One known issue i observed is that an app called start11 could modify the taskbar window causing further instability, but this would be subjected case-by-case to any third party plugins and their implementation, and would be abit of a pain to just create a case for each one of them.

I think flipping the flag back to false after 8 seconds would mean that this PR essentially increases the 1000ms fixed wait time to 8000ms wait time, but allow wait to end earlier if the taskbar is ready; OR if never flipping it to false, then the taskbar features are never loaded again, maybe consider exit the program or showing a message telling the user that taskbar is buggy? 

The easier solution would be yes, flipping it back to false, I think 8 seconds is more than enough for any heavy loading, but I think the ultimate decision of this would be how you imagine the app behave.

What do you think? I'll finalise the relevant changes in the next commit, depending on what we decide here.

mclt0568:
Or, keeping it as true, and maybe the next explorer restart externally (by the buggy plugin, or by the user manually) solves the buggy taskbar, then the taskbar window can reload properly, but never actively triggers a explorer restart by force loading TaskbarWindow, I think would be a better option?

In this case since it would be handled again when this app receives `WM_TASKBARCREATED`, so nothing needs to be changed in the current code i think

unchihugo:
Thanks for the quick implementation :)

I guess your points make sense! In that case we can keep it set to true, but for edge-case scenarios where the taskbar might take longer than 8 seconds to restart (like really slow computers), perhaps we should extend the timeout to a minute (maybe with exponential backoff if it's resource intensive?), what do you think?

EDIT: I also noticed that there's a duplicate import and the formatting could be fixed up just a little bit. Are you able to look into that as well? thanks!

mclt0568:
sorry for the messy code 😭, should be fixed now

i think exponential backoff would be a great idea, i might get it in later, its getting late today, for now its 1 minute timeout

unchihugo:
Hi @mclt0568, just wondering if you're still planning to finish up this implementation :)

PR #321, opened 2026-01-29
Add option to exclude volume keys from triggering media flyout

neegool:
Users might not want to consider volume up/down/mute as trigger for the media flyout. This PR adds a switch to toggle exclusion.


unchihugo (changes requested):
Hi, thank you for the PR, looks great!

Before we merge, I actually think that this new toggle should be lower in the page (similar to the Insert toggle in Lock Keys), as it's more a configuration option rather than a customization option. What do you think?

neegool:
> Hi, thank you for the PR, looks great!
> 
> Before we merge, I actually think that this new toggle should be lower in the page (similar to the Insert toggle in Lock Keys), as it's more a configuration option rather than a customization option. What do you think?

Sure thing, let's put it at the bottom of the list (below `Pause Other Media Automatically`)

Issue #373, opened 2026-02-09
[BUG] Taskbar Visualizer


Raffi67890:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my bug
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my issue

### Describe the bug

The display only starts when you deactivate the option in the settings and reactivate it again. When playback is active, you turn off the PC or, with mobile devices in the energy saving mode, the display remains off

Additionally, the sensitivity control doesn't seem to make a difference

### To Reproduce

1. Start a Playback over Spotify
2. Go to FluentFlyout settings / Taskbar Visualizer / Disable Taskbar Visualizer / Enable Taskbar Visualizer
3. The Visualizer Starts 
4. Stop The Playback 
5. Turn Off the PC ore set the Pc to Energy Saveing 
6. wait like 1 too 5 min
7. Start up the PC 
8. Start up The Playback
9. Visualizer Dont work.

### Desktop

- OS: [e.g. Windows 11] 25H2
- FluentFlyout version: [e.g. v2.7.0]

unchihugo:
Hi, thanks for the report! are you able to provide the logs so I can check what's wrong? Thanks!

Raffi67890:
[logs.2026-02-06 09-25-41.txt](https://github.com/user-attachments/files/25204026/logs.2026-02-06.09-25-41.txt)
[logs.2026-02-09 09-23-25.txt](https://github.com/user-attachments/files/25204025/logs.2026-02-09.09-23-25.txt)
[logs.2026-02-09 09-39-50.txt](https://github.com/user-attachments/files/25204022/logs.2026-02-09.09-39-50.txt)
[logs.2026-02-10 07-04-27.txt](https://github.com/user-attachments/files/25204024/logs.2026-02-10.07-04-27.txt)
[settings.xml](https://github.com/user-attachments/files/25204023/settings.xml)

I hope these are the ones you need because the logs button just throws me at documents but well I have them now from appdata

and I'm sorry if my English is a bit confusing at times. I'm writing using Google Translate because I'm from Germany

unchihugo:
Should be fixed in the latest versions.

Issue #358, opened 2026-02-07
[FEATURE REQUEST] Device selection for Taskbar Visualizer


DmitriySalnikov:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my feature request
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my request

### Description

I have been using virtual audio cables for several years to split the sound into multiple tracks during recording. I use `Voicemeeter` and its `Cable` + `Cable A`. All sound from browsers and players is output to `Cable A`.
And when `Taskbar Visualizer` attempts to capture audio, it cannot obtain any audio for visualization (except for system sounds that have not been redirected to another V.A.C.).

Therefore, it would be useful to be able to change the capture device in the `FluentFlyout` settings.

### Additional context

To work around this issue, we can use system settings (`Settings > System > Sound > Volume Mixer`) to change the default output device for `FluentFlyout`, as well as for the playback application. Then, we need to restart the visualizer through the settings.

https://github.com/user-attachments/assets/db2c0848-594b-453e-b591-34b63dc02720

PR #379, opened 2026-02-10
Fix visualizer always listening to the first master output device

### LINKED ISSUE #367: [BUG] Taskbar visualiser not showing up

### Describe the bug

As per title, nothing shows up when playing from web neither from foobar. I tried with both widgets set in every position and, when setting the visualiser towards the sides of the screen, only an empty space appears.

Oh, Fluentflyout doesn't work with VLC - i do not use it, played for testing purposes.

### To Reproduce

1.set main widget
2.set addon widget
3.play music
...

_No response_

### Desktop

- OS: Windows 11 25h2 26220.7752
- FluentFlyout version: 2.7.0

unchihugo:
Fixed visualizer always capturing to the first master output device by listening to master output device changes and changing the capture accordingly. Fixes #367, potentially #358 and #373 but needs further investigation

### DIFF
--- a/FluentFlyoutWPF/Classes/AudioDeviceMonitor.cs
+++ b/FluentFlyoutWPF/Classes/AudioDeviceMonitor.cs
@@ -89,6 +89,7 @@ namespace FluentFlyoutWPF.Classes
                 {
                     Logger.Error(ex, "Failed to unregister device notification callback");
                 }
+                _deviceEnumerator.Dispose();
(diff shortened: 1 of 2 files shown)


DmitriySalnikov:
Selecting and automatically switching to the current default device works without any problems (at first glance).

However, if you do not plan to add device selection to the settings window, I think it would be worth adding a note such as "use system settings to select the audio output device".

Issue #350, opened 2026-02-05
[BUG] Flyout and Widget on wrong target display


thuglifesk:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my bug
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my issue

### Describe the bug

Flyout and Widget appearing on the wrong screen. I have selected target display 1 in options but it keeps appearing on screen 2. If I select screen 2 it appears on screen 1 until reboot. 

### To Reproduce

1.
2.
3.
...

### Desktop

- OS: Win 11
- FluentFlyout version: 2.6.1

PR #404, opened 2026-02-22
Improved the monitor/display ordering in settings.

Hikaru11747:
Fixing for #350 

- Ordered the monitor list, so that the first element will always be the primary monitor, then follows a left to right order. 
- Merged the UpdateMonitorList() from SystemPage and TaskbarWidgetPage using a MonitorUtil static class.(Maybe a better sol would be to create a view model to update the list dynamically?)

### DIFF
--- a/FluentFlyoutWPF/Classes/NativeMethods.cs
+++ b/FluentFlyoutWPF/Classes/NativeMethods.cs
@@ -3,7 +3,6 @@
 
 using System.Runtime.InteropServices;
 using System.Text;
-using Windows.Devices.Display.Core;
 
 namespace FluentFlyout.Classes;
 
--- a/FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
+++ b/FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
@@ -1,12 +1,11 @@
 ﻿// Copyright © 2024-2026 The FluentFlyout Authors
 // SPDX-License-Identifier: GPL-3.0-or-later
 
+using FluentFlyout.Classes;
 using System.Runtime.InteropServices;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Interop;
-
-using FluentFlyout.Classes;
 using static FluentFlyout.Classes.NativeMethods;
 
 namespace FluentFlyoutWPF.Classes.Utils;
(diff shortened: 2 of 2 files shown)


unchihugo:
Hi, thanks for contributing @Hikaru11747!

The consolidation of duplicated code looks good, thank you for handling that!

I have some thoughts after reading the code:
1. I believe that currently the numbering is based off of how Windows enumerates the monitors itself (in Windows display settings) if I remember correctly. By changing the way it's enumerated, I believe might go against standards Windows has set, right?
2. Since we have a new util class for monitors now, do you think we should move the existing monitor methods in WindowHelper to it as well?

Now that I think about it, it might be better to display the names of the monitors in brackets next to the number (like how Windows display settings does) instead of reordering the monitors. Names could be stored in another variable then returned after the number when setting the ComboBox (so it doesn't save to the settings). What do you think?

Hikaru11747:
Hi @unchihugo, 
1. I observed the list that we get from EnumDisplayMonitors is not ordered properly, I see my primary monitor as the 2nd element in the list, which is confusing in the combo boxes, So the order in the settings and the list we get is mismatching, It would be great if it follows how the display settings does, Thinking how to make it respect that.
2. Yes it would be great to migrate the windows helper funcs to the util class.

We can add the monitor names for more clarity for sure!

Hikaru11747:
https://learn.microsoft.com/en-us/answers/questions/1572062/how-to-get-the-monitor-number-displayed-in-the-sys See this.

unchihugo:
> https://learn.microsoft.com/en-us/answers/questions/1572062/how-to-get-the-monitor-number-displayed-in-the-sys See this.

Oh, interesting! In that case, we don't have to make it match what windows shows. I think names will suffice.

FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
```diff
@@ -0,0 +1,148 @@
...
+            .ToList();
+    }
+    private static string GetMonitorFriendlyName(string deviceId)
+    {
+        var displayDevice = new NativeMethods.DISPLAY_DEVICE
+        {
+            cb = Marshal.SizeOf<NativeMethods.DISPLAY_DEVICE>()
+        };
+
+        // Step 1: Get adapter info (NULL, iDevNum)
```
unchihugo: These comments are a bit confusing, could you modify them?

FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
```diff
@@ -0,0 +1,148 @@
+﻿// Copyright © 2024-2026 The FluentFlyout Authors
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+using System.Runtime.InteropServices;
```
unchihugo, line 5: Usings can be sorted and cleaned by using `CTRL + R, CTRL + G`

FluentFlyoutWPF/Classes/NativeMethods.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Runtime.InteropServices;
 using System.Text;
+using Windows.Devices.Display.Core;
```
unchihugo: This is an unused `using`

unchihugo (changes requested):
Looks good now! I haven't encountered any issues during testing, so it seems ready. Could you go over the comments I've added before we merge? Thanks!

EDIT: btw, have you noticed that this fixed the issue you've tagged?

Hikaru11747:
Hey @unchihugo, I've made the requested changes, please have a look! And I can confirm it fixes #350

Issue #336, opened 2026-02-01
[FEATURE REQUEST] Add Import/Export functionality for settings and configurations


suheandzl:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my feature request
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my request

### Description

Hi there! I am a big fan of this software and have already purchased it through the Microsoft Store to support the development. Thank you for your continuous efforts in improving this tool!

I would like to request a new feature: the ability to Import and Export software settings and configurations

An Import/Export feature would provide peace of mind and significantly improve the user experience for power users.

Thank you again for your time and for this excellent software!

Because I used AI translation, I hope my meaning is accurately expressed.

unchihugo:
Hi, thanks for the suggestion! This is a great idea. 

For now, you can find the settings file in the same location as the logs folder (located at `C:\Users\USERNAME\AppData\Local\Packages\unchihugo.FluentFlyout_69b7b6qge1ahj\LocalCache\Roaming\FluentFlyout` if you have the Microsoft Store version)!

suheandzl:
Hello, thank you for your reply! I will manually back up the software configurations before this feature is implemented. Thank you!

FaroJoaoFaro:
Hey! I'm not sure if my implementation is the best, nor if the placement of the buttons is the best but I believe it's all done and working, at least from my testing. I hope it could be of good use. I'm loving the project! Keep it up

PR #424, opened 2026-02-27
Add Import/Export functionality for settings and configurations

FaroJoaoFaro:
This intends to add the functionality of import and export settings file as described and requested on issue #336


FaroJoaoFaro:
1- You were completely right! Totally usable; Thanks! 
2- Yeah I didn't know we didn't need those conditions. I saw them in the same file above and that's why I used them also. I now removed all of those.
(previous ones, not mines, were on lines 45 and 86 on 'FluentFlyoutWPF/Pages/HomePage.xaml.cs'

Hope this is good!

PR #435, opened 2026-03-02
Added a setting for selecting the side to display media buttons on the widget (left or right)

mak7im01:
- Useful for those who use the right-hand widget

https://github.com/user-attachments/assets/9fab62df-8eea-4885-b73f-c13c5c242c09

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -75,7 +75,7 @@ public partial class MainWindow : MicaWindow
     private LockWindow? lockWindow;
     private DateTime _lastSelfUpdateTimestamp = DateTime.MinValue;
 
-    private TaskbarWindow? taskbarWindow;
+    internal TaskbarWindow? taskbarWindow;
 
     internal static volatile bool ExplorerRestarting = false;
 
@@ -454,9 +454,6 @@ public partial class MainWindow : MicaWindow
 
     public void UpdateTaskbar()
     {
-        // Reorder controls if needed
-        taskbarWindow?.Widget?.ReorderControls();
-        
         if (!mediaManager.IsStarted || mediaManager.GetFocusedSession() == null)
         {
             taskbarWindow?.UpdateUi("-", "-", null, GlobalSystemMediaTransportControlsSessionPlaybackStatus.Closed);
--- a/FluentFlyoutWPF/ViewModels/UserSettings.cs
+++ b/FluentFlyoutWPF/ViewModels/UserSettings.cs
@@ -677,7 +677,9 @@ public partial class UserSettings : ObservableObject
     partial void OnTaskbarWidgetControlsPositionChanged(int oldValue, int newValue)
     {
         if (oldValue == newValue || _initializing) return;
-        UpdateTaskbar();
+        
+        MainWindow mainWindow = (MainWindow)Application.Current.MainWindow;
+        mainWindow.taskbarWindow?.Widget?.ReorderControls();
     }
 
     partial void OnTaskbarVisualizerPositionChanged(int oldValue, int newValue)


mak7im01:
> I've got a few requests looking at the code:
> 3. Just another thought but not really important; do you think these settings cards should go in a collapsible card instead (like how the padding settings are bundled) or keep it as is?

I tried to fulfill all the requests.

Regarding this, I think we can leave it as is. If another item is added, then we can place these settings cards in a collapsible card.

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml
```diff
@@ -49,22 +65,6 @@
                             <TextBlock Name="SongArtist" Text="" FontWeight="Normal" Opacity="0.5" Margin="0,-1.5,0,0" TextTrimming="CharacterEllipsis" Foreground="{ui:ThemeResource TextFillColorPrimaryBrush}" />
                         </StackPanel>
 
-                        <StackPanel Name="ControlsStackPanel" Orientation="Horizontal" VerticalAlignment="Center" Width="Auto" Margin="8,0,0,0">
```
unchihugo, line 52: The original `ControlsStackPanel` should be kept instead of the new one - reason is because `Right` position is the default, so it's easier for maintainability!

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -43,6 +43,15 @@ public TaskbarWidgetControl()
         // Set DataContext for bindings
         DataContext = SettingsManager.Current;
 
+        // Subscribe to position changes to reorder controls
```
unchihugo: You can listen to changes in `UserSettings.OnTaskbarWidgetControlsPositionChanged()` instead - this will be much more efficient than firing on every property change! I see you've already created a method for it in UserSettings, just change `UpdateTaskbar()` to the reorder method :)

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -57,6 +66,42 @@ public TaskbarWidgetControl()
...
 
         Background = new SolidColorBrush(Color.FromArgb(1, 0, 0, 0)); ;
+        
+        // Initialize control order
+        ReorderControls();
+    }
+
+    private void ReorderControls()
+    {
+        // Remove all children from MainStackPanel
```
unchihugo: I've quickly tested this myself, and I believe replacing this whole function with the following should yield the same results!

``` c#
MainStackPanel.Children.Remove(ControlsStackPanel);

// Reorder based on position setting
if (SettingsManager.Current.TaskbarWidgetControlsPosition == 0)
{
    // Left: Controls, Image, Info
    MainStackPanel.Children.Insert(0, ControlsStackPanel);
    ControlsStackPanel.Margin = new Thickness(2, 0, 6, 0); // for some reason margins are weird on left side
}
else
{
    // Right: Image, Info, Controls
    MainStackPanel.Children.Add(ControlsStackPanel);
    ControlsStackPanel.Margin = new Thickness(8, 0, 0, 0);
}
```

unchihugo (changes requested):
@mak7im01 this is a great step forward! I personally didn't even know that you were able to reorganize stackpanels like that.

Before we merge, there's a few more things to take care of - I'll list them as comments. Thank you for your contributions so far! 

Btw, you don't have to squash all your commits, I'm fine with multiple commits if they each are valuable!

mak7im01:
> Btw, you don't have to squash all your commits, I'm fine with multiple commits if they each are valuable!

If you need to merge commits, write

unchihugo:
@mak7im01 Are you able to just call ReorderControls() from UserSettings instead of adding it to UpdateTaskbar()?

Issue #328, opened 2026-01-30
[FEATURE REQUEST] Pause/Play when clicking on the track cover in the widget


mak7im01:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my feature request
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my request

### Description

Hello. Please add a "Pause/Play" feature when clicking on a track's cover in the widget.

### Additional context

![Image](https://github.com/user-attachments/assets/4c987791-ec79-4284-aa53-8386e5f25833)

unchihugo:
Hi, thanks for the suggestion! Just wondering, why would you prefer clicking the cover icon to play/pause instead of the adding controls option?

mak7im01:
I don't use the add controls option. I rarely use the previous/next controls, but I do use the Pause/Play button quite often. I don't want to use the add controls option for this purpose, as it takes up a lot more space, and I only need one button.
The most important thing is that when you pause, the corresponding icon appears, making it so tempting to click it to play.

![Image](https://github.com/user-attachments/assets/b1f552e8-87fa-414d-ab23-9fdec67be4eb)

Hikaru11747:
Since we have a controls option, I was thinking that when controls are disabled, we could use the track cover as a Play/Pause toggle. If the user enables the controls option, this behavior would be disabled.

Hover over the track cover when music is playing shows the pause button, the play button will be constantly be shown when paused until played.

PR #442, opened 2026-03-04
Added Pause/Play when clicking on a track's cover in the widget

mak7im01:
- Implemented the feature from this request https://github.com/unchihugo/FluentFlyout/issues/328

- Added a switch to Media Flyout

https://github.com/user-attachments/assets/3cdd638f-3ab2-4d63-8b98-da341821eba6

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 	<img alt="GitHub contributors" src="https://img.shields.io/github/contributors-anon/unchihugo/fluentflyout?labelColor=midnightblue&color=goldenrod">
 </p>
 <p align="center">
-  <strong>English</strong> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.zh.md">简体中文</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.nl.md">Nederlands</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.tr.md">Türkçe</a>
+  <strong>English</strong> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.zh.md">简体中文</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.nl.md">Nederlands</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.tr.md">Türkçe</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.ru.md">Русский</a>
 </p>
 
 ---
--- a/README.nl.md
+++ b/README.nl.md
@@ -8,7 +8,7 @@
 	<img alt="GitHub contributors" src="https://img.shields.io/github/contributors-anon/unchihugo/fluentflyout?labelColor=midnightblue&color=goldenrod">
 </p>
 <p align="center">
-  <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.md">English</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.zh.md">简体中文</a> | <strong>Nederlands</strong> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.tr.md">Türkçe</a>
+  <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.md">English</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.zh.md">简体中文</a> | <strong>Nederlands</strong> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.tr.md">Türkçe</a> | <a href="https://github.com/unchihugo/FluentFlyout/blob/master/README.ru.md">Русский</a>
 </p>
 
 ---
--- a/README.tr.md
+++ b/README.tr.md
@@ -8,7 +8,7 @@
 	<img alt="GitHub contributors" src="https://img.shields.io/github/contributors-anon/unchihugo/fluentflyout?labelColor=midnightblue&color=goldenrod">
 </p>
 <p align="center">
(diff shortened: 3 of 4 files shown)


unchihugo:
Hi @mak7im01, thank you for the implementation! Since we already have the Widget Media Controls option for playback controls, I'm hesitant to add this. Clicking album art isn't a pattern that feels native to Windows widgets, and one of our goals is to stay close to that native feel. I appreciate your continuous help regardless, and I'm happy to discuss!

Hikaru11747:
Just to digress, let's take a look at Apple Music for Windows, which has a similar design choice.

Unhovered.
image

Hovered (Prompts to mini player.)
image

So as we had a discussed in #328 the album art imo needs to have a play icon when the media is paused always and only show pause icon when hovered over the album art.

Additionally, we could have a dropdown for the controls instead of two toggles that fight with each other.

mak7im01:
Hi @unchihugo This feature was added for those who don't use media buttons (it intentionally doesn't work with enabled media buttons). As I wrote in the proposal, the "Pause" button hints that clicking it will start playback (which is logical).
This feature doesn't change the app's logic, but merely expands its capabilities. Users who don't need it can simply disable it in the program settings, which would be natural for Windows widgets (as you wrote).
In my opinion, the more customization an app offers, the better. This allows everyone to customize the app and use it however they prefer. It doesn't force you to use a given feature.

@Hikaru11747 Judging by your screenshots, I prefer the way it's implemented/looks in FluentFlyout.

unchihugo:
Honestly @Hikaru11747 that sounds like a good idea. We can have hover states (like how Apple did it) and replace the setting with a dropdown instead to declutter the settings. If it works for the design-first company, it must be good :) We can also hide the left or right controls position depending on whether the button controls are selected. 

EDIT: sorry for the PR close, I had accidentally pressed the button on mobile

mak7im01:
> So as we had a discussed in #328 the album art imo needs to have a play icon when the media is paused always and only show pause icon when hovered over the album art.

After adding "Added pause/play functions when clicking on the track cover in the widget" I wanted to start with implementing this function

mak7im01:
Second option:

https://github.com/user-attachments/assets/06dd5940-ee45-4627-aa9f-bd0405478cb6

unchihugo:
> Second option:
> 
>  Video_2.mp4

That looks great! Could you make the button span the whole cover image size? It looks like it's a bit too small in its current style. 

Also before we continue, it's important that you keep the hover states on the MainBorder. By changing the hover states to only the song info, we move away from the native Windows taskbar's styling. The SongImage_MouseLeftButtonUp() code also looks quite messy, and I'm sure optimizations could be done to avoid the if statements, like disabling hit checks instead.

mak7im01:
@unchihugo The second option is a completely different project, specifically the second option, so it needs to be integrated into the first option.

I understand about the button size.

I'd like to describe the second option (if I understood you correctly):
1. With media buttons disabled, hovering the mouse over the entire MainBorder block would display the track's status on the cover (pause/play), and clicking on the cover would trigger the pause/play action.
2. I would remove the setting I added so that when the media buttons are disabled, what I described above would happen. This would clearly indicate to the user that clicking on the cover triggers pause/play. And when the media buttons are enabled, there would be nothing on the cover, since the track's status is visible on the buttons themselves (the media buttons).

The question remains:
For point 2 (2), should the cover be active so that clicking on it opens the control window, or only when clicking on the track title?

mak7im01:
@unchihugo Should I implement what I described above?

unchihugo:
The comment is not really addressing my problem, which is that by removing the MainBorder hover states, we're drifting away from native-like implementation: the new hover states ignore the whole container, unlike how it was previously. The best solution here is to revert those changes and implement new code in the current hover methods. Of course, the cover hovers can be kept.

Regarding the cover image displaying an icon; this could be activated when hovering over the cover, like in your demo video. It should also indeed be an option instead of default behavior, because buttons inside containers on the taskbar isn't very native unfortunately.

Regarding your last question: similarly to the current existing control buttons, this doesn't have to show the media flyout.

mak7im01:
Can you rephrase that? I'm not quite sure what I need to do to Play/Pause  the song, was it added to the cover in the widget?

unchihugo:
> Can you rephrase that? I'm not quite sure what I need to do to Play/Pause the song, was it added to the cover in the widget?

1. Edits to the MainGrid hover states have to be reverted. This new functionality is impractical.

2. You mentioned that the cover play/pause would display if you hover the container - it should only display when you hover the cover image.

3. An individual setting should be kept for the cover play/pause functionality because it's not really a taskbar native thing, but could be adjusted later.

4. Your last question was whether to display the media flyout when clicking the cover play/pause: it shouldn't display the media flyout (similarly to the current controls functionality).

mak7im01:
I reworked the code, but haven't uploaded it yet because I still can't figure out from the text what it should look like.
Right now, it looks like this:

https://github.com/user-attachments/assets/5385a376-b9fe-4469-89cd-04a4de269be2

unchihugo:
Looks great so far!

> I still can't figure out from the text what it should look like.

Perhaps this could work:
**"Play/Pause via Album Cover"**
"Click the album cover to play or pause media"

mak7im01:
> Perhaps this could work: **"Play/Pause via Album Cover"**

In the video demonstration, this is exactly what is happening now

unchihugo:
Oh sorry, thought you meant to reword the setting. What exactly do you need to know in more detail?

mak7im01:
Is what was demonstrated in the last video suitable? And if anything is wrong, please let me know what changes should be made

unchihugo:
It looks good from the video, though some of the things I've mentioned will need to be checked through the code. Could you also rephrase the setting to my previous comment?

mak7im01:
I'll close this thread for now. I'll reopen it when the code is ready.

Issue #448, opened 2026-03-05
[FEATURE REQUEST] Mouse cursor aware Lock Flyout.


Hikaru11747:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my feature request
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my request

### Description

I noticed in my dual monitor setup that the lock flyout appears on the selected display based on the settings, And I suggest to have toggle which will allow to follow the mouse cursor i.e determine the display the cursor is in and show the lock flyout on that display.

### Additional context
Similar to how Command Palette from power toys works.

PR #449, opened 2026-03-05
Added a toggle option to select monitor with cursor for the lock flyout.

Hikaru11747:
Addresses #448.
Select the monitor with the mouse cursor to show the flyout.

### DIFF
--- a/FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
+++ b/FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
@@ -24,13 +24,13 @@ public static class MonitorUtil
     }
 
 
-    public static MonitorInfo getSelectedMonitor(int index = 0)
+    public static MonitorInfo GetSelectedMonitor(int index = 0)
     {
         var monitors = GetMonitors();
         return monitors[Math.Clamp(index, 0, monitors.Count - 1)];
     }
 
-    private static MonitorInfo getMonitorInfoInternal(IntPtr hMonitor)
+    private static MonitorInfo GetMonitorInfoInternal(IntPtr hMonitor)
     {
         var info = new NativeMethods.MONITORINFOEX();
         info.cbSize = Marshal.SizeOf<NativeMethods.MONITORINFOEX>();
@@ -67,7 +67,7 @@ public static class MonitorUtil
     public static MonitorInfo GetMonitor(IntPtr hwnd, MonitorFromWindowFlags flag = MonitorFromWindowFlags.DEFAULTTONEAREST)
     {
         var hMonitor = MonitorFromWindow(hwnd, (int)flag);
-        return getMonitorInfoInternal(hMonitor);
+        return GetMonitorInfoInternal(hMonitor);
     }
 
     public static MonitorInfo GetMonitor(Window window, MonitorFromWindowFlags flag = MonitorFromWindowFlags.DEFAULTTONEAREST)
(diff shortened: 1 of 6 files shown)


FluentFlyoutWPF/Classes/Utils/MonitorUtil.cs
```diff
@@ -143,4 +143,28 @@ public static void UpdateMonitorList(ComboBox comboBox, Func<int> getSelectedInd
         comboBox.SelectedIndex = selectedMonitor;
         setSelectedIndex(selectedMonitor);
     }
+
+    public static MonitorInfo getActiveCursorMonitor()
```
unchihugo: The two new monitor methods are named quite differently from each other (getActiveMonitorCursor() and getMonitorWithFocusedWindows() *also windows should probably be singular). We should rename either of them to match each other.

FluentFlyoutWPF/ViewModels/UserSettings.cs
```diff
@@ -244,6 +244,13 @@ public string LockKeysDurationText
     [XmlElement(ElementName = "LockKeysBoldUI")]
     public partial bool LockKeysBoldUi { get; set; }
 
+    /// <summary>
+    /// Use If multiple monitors are in use, shows the flyout on the monitor with the cursor.
+    /// </summary>
+    [ObservableProperty]
+    [XmlElement(ElementName = "LockKeysBoldUI")]
```
unchihugo: The XmlElement rename should be removed, and we should rename "LockKeysPositionPreference" with LockKeysMonitorPreference (to match other monitor settings), and assign the default = 0 in the initialization

FluentFlyoutWPF/Windows/LockWindow.xaml.cs
```diff
@@ -116,4 +120,19 @@ public async void ShowLockFlyout(string? key, bool isOn)
             // do nothing
         }
     }
+
+    private MonitorInfo selectLockPreferedMonitor()
```
unchihugo: The naming of this method is a little confusing, are you able to rename this as well?

FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
```diff
@@ -121,6 +121,11 @@ Note: As an open-source project, these features are also available for free via
     <System:String x:Key="LockKeysStayDurationDescription">(default: 2000 ms)</System:String>
     <System:String x:Key="LockKeysBoldUITitle">Bold Symbol and Font</System:String>
     <System:String x:Key="LockKeysBoldUIDescription">Emphasizes the flyout symbol and font</System:String>
+    <System:String x:Key="LockKeysCursorUITitle">Prefered monitor position</System:String>
```
unchihugo: We should rename this to "Preferred Target Display" (matching existing monitor settings) and options "Default (Flyout Target Display)", "Monitor with focused window" and "Monitor with cursor"

unchihugo (changes requested):
Looks good so far @Hikaru11747! Implementation seems great. I've just read the code and noted some changes that are needed before we merge.

Hikaru11747:
Ah missed that! my bad, I'll make those changes

unchihugo:
Oh, these suggestions are definitely worth checking out before we merge! There's some breaking issues here.

Hikaru11747:
Yep, Let me address those

Issue #354, opened 2026-02-07
[FEATURE REQUEST or BUG, whatever this is] Add flyout support for Headphone volume controls


WinDogeNT:
### Confirm that you have done the following:

- [x] I have searched that there is no existing issue regarding my feature request
- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my request

### Description

Thanks for a good app, I really like it!
However, there is this lack of feature or bug that when I press a volume button or pause button on my headphones (example: JBL Tune), the flyouts doesn't work like normally pressing on the function keys. I'd hope devs implement these.

jh30uk:
It started doing the same for me a few builds ago (weeks/month), I did report it but there is a temp fix for now but it will not work after you reboot your PC.

Check the new setting to ignore volume buttons and fully close the App then reopen the App and uncheck the same setting and it will work for the time being.

Also the visualizer does nor remember if I change it to 15 bars etc, it reverts  to default after a restart of the App or PC.

Issue #286, opened 2026-01-14
[FEATURE REQUEST] audio flyouts for headphone buttons


WinDogeNT:
I usually use my BT headphones when listening to music on my PC. My headphones have buttons that I can use to control volume or play/pause, but FluentFlyouts doesn't support those. It would be nice if I see a FluentFlyouts media flyout when I use my headphones to turn up my volume, instead of having to press F3 for the flyout to show.

Issue #147, opened 2025-11-11
[Touch-Pad Gesture Incompatibility]


sardam3:
**Describe the bug**
The media flyout does not appear when using the Windows 11 "Change audio and volume" four-finger touchpad gesture. Instead, the default Windows 11 volume flyout is displayed. The custom flyout works as expected when using keyboard volume shortcuts, but not with the touchpad gesture.

**To Reproduce**
1. Ensure the flyout application is running.
2. In Windows 11, go to Settings > Bluetooth & devices > Touchpad > Four-finger gestures.
3. Set one of the gestures (e.g., "Swipes") to "Change audio and volume".
4. Put on a music from a supported player (I was using Spotify).
5. Perform that four-finger gesture on the touchpad to change the volume or skip a track.
6. See error: The default Windows 11 volume flyout appears, not the custom application's flyout.

**Expected behavior**
When using the four-finger touchpad gesture to change volume, the custom application's media flyout should appear, overriding the system default (just as it does for keyboard shortcuts).

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version v2.1.0

**Additional context**
This issue seems to be specific to the event fired by the built-in Windows 11 touchpad gesture. The application correctly intercepts standard keyboard media keys (Volume Up, Volume Down, Mute), but it fails to intercept this particular touchpad gesture event.

Issue #138, opened 2025-11-01
[BUG] touchpad media gestures


xBesoz:
**Describe the bug**
When I stop or play music with the touchpad gestures it doesn't show the flyout it only works the the function buttons on my keyboard.

Windows 11.

unchihugo:
Hi, as far as I know there's no way to detect media control on Windows other than the keyboard inputs. That means Bluetooth controls and touchpad gestures cannot be detected.

I'll leave this issue open in case anyone has any idea on how to make this work. It would be greatly appreciated!

PR #466, opened 2026-03-10
[Feature] NAudio For Volume Changed Instand Of Keyboard Hook

kitUIN:
### New Feature Support

- Separate the keyboard hook part from the main window and make it an independent service: `InputMonitorService`. This service notifies the main program through events. And the new mode of NAudio is also included in this service.
	- `VolumeChangedEvent`: mute / volume up / volume down
	- `MediaKeyPressedEvent`: play/pause, next, previous, stop
	- `LockKeyPressedEvent`: CapsLock / NumLock / ScrollLock / Insert
- Due to the changes in the volume of the listening system, it will support all sources that cause volume changes, such as the touchpad, headphones, etc.
- Added setting options. The default mode is NAudio. If there are compatibility issues, the previous Keyboard Hook mode can be selected.

### Related issues
  - https://github.com/unchihugo/FluentFlyout/issues/354
  - https://github.com/unchihugo/FluentFlyout/issues/286
  - https://github.com/unchihugo/FluentFlyout/issues/147
  - https://github.com/unchihugo/FluentFlyout/issues/138
  - https://github.com/unchihugo/FluentFlyout/issues/79
  - https://github.com/unchihugo/FluentFlyout/issues/54
  - https://github.com/unchihugo/FluentFlyout/issues/18

### DIFF
--- a/FluentFlyoutWPF/Services/InputMonitorService.NAudio.cs
+++ b/FluentFlyoutWPF/Services/InputMonitorService.NAudio.cs
@@ -1,6 +1,7 @@
 // Copyright © 2024-2026 The FluentFlyout Authors
 // SPDX-License-Identifier: GPL-3.0-or-later
 
+using System.Runtime.InteropServices;
 using FluentFlyout.Classes.Settings;
 using FluentFlyoutWPF.Classes;
 using FluentFlyoutWPF.Services.Args;
@@ -15,7 +16,10 @@ public sealed partial class InputMonitorService
     private string? _nAudioRenderDeviceId;
     private bool _nAudioDeviceChangeSubscribed;
 
-    
+
+    /// <summary>
+    /// Rebinds On Default Render Device Changed
+    /// </summary>
     private void OnDefaultRenderDeviceChanged(object? sender, DefaultDeviceChangedEventArgs e)
     {
         if (e.DataFlow != DataFlow.Render)
@@ -31,7 +35,7 @@ public sealed partial class InputMonitorService
 
         _ = Task.Run(BindNRenderDevice);
     }
-    
+
     /// <summary>
     /// Starts NAudio-based monitoring by subscribing to default render device changes
     /// and binding volume notifications from the current default render device.
@@ -79,7 +83,8 @@ public sealed partial class InputMonitorService
         }
 
         string nextDeviceId = nextDevice.ID;
-        if (_nAudioRenderDevice != null && string.Equals(_nAudioRenderDeviceId, nextDeviceId, StringComparison.OrdinalIgnoreCase))
+        if (_nAudioRenderDevice != null &&
+            string.Equals(_nAudioRenderDeviceId, nextDeviceId, StringComparison.OrdinalIgnoreCase))
         {
             nextDevice.Dispose();
             return true;
@@ -89,7 +94,17 @@ public sealed partial class InputMonitorService
 
         _nAudioRenderDevice = nextDevice;
         _nAudioRenderDeviceId = nextDeviceId;
-        _nAudioRenderDevice.AudioEndpointVolume.OnVolumeNotification += OnNVolumeNotification;
+        try
+        {
+            _nAudioRenderDevice.AudioEndpointVolume.OnVolumeNotification += OnNVolumeNotification;
+        }
+        catch (InvalidComObjectException)
+        {
+            // The audio device failed during the binding process.
+            Logger.Warn("Binding NAudio render device failed about COM: {}", nextDeviceId);
+            return false;
+        }
+
         return true;
     }
 
@@ -98,16 +113,44 @@ public sealed partial class InputMonitorService
     /// </summary>
     private void UnbindNRenderDevice()
     {
+        _nAudioRenderDevice = null;
+        _nAudioRenderDeviceId = null;
+
         if (_nAudioRenderDevice == null)
         {
             return;
         }
 
-        _nAudioRenderDevice.AudioEndpointVolume.OnVolumeNotification -= OnNVolumeNotification;
-        _nAudioRenderDevice.Dispose();
-        _nAudioRenderDevice = null;
-        _nAudioRenderDeviceId = null;
+        try
+        {
+            try
+            {
+                _nAudioRenderDevice.AudioEndpointVolume.OnVolumeNotification -= OnNVolumeNotification;
+            }
+            catch (InvalidComObjectException)
+            {
+                // ignore
+            }
+
+            try
+            {
+                _nAudioRenderDevice.Dispose();
+            }
+            catch (InvalidComObjectException)
+            {
+                // ignore
+            }
+        }
+        catch (Exception)
+        {
+            // ignore
+        }
+        finally
+        {
+            _nAudioRenderDevice = null;
+        }
     }
+
 
     /// <summary>
     /// Handles NAudio endpoint volume notifications and dispatches a volume change event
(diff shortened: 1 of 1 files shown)


FluentFlyoutWPF/Services/InputMonitorService.cs
```diff
@@ -0,0 +1,326 @@
...
+    private readonly Thread _dispatchThread;
+
+    private IntPtr _hookId = IntPtr.Zero;
+    private LowLevelKeyboardProc? _hookProc;
+    private long _lastFlyoutTime;
+    private long _lastLockKeyEventTime;
+
+    private bool _isStarted;
+
+    private const int MEDIA_KEY_PLAY_PAUSE = 0xB3;
```
unchihugo: Do you think we should add these constants to `NativeMethods.cs` or keep them here?

FluentFlyoutWPF/Services/InputMonitorService.cs
```diff
@@ -0,0 +1,326 @@
...
+            
+            if ((currentTime - lastFlyoutTime) >= 500)
+            {
+                Interlocked.Exchange(ref _lastFlyoutTime, currentTime);
+
+                if (volumeKeyPressed)
+                {
+                    DispatchEventAsync(DispatchVolumeChanged);
+                }
+                else // 此时必然是 mediaKeyPressed 为 true
```
unchihugo: Could you translate this comment as well?

FluentFlyoutWPF/Services/InputMonitorService.cs
```diff
@@ -0,0 +1,326 @@
...
+                break;
+            case VK_INSERT:
+                RaiseLockKeyPressed(LockKeyType.INSERT, Keyboard.IsKeyToggled(Key.Insert));
+                break;
+        }
+
+        return CallNextHookEx(_hookId, nCode, wParam, lParam);
+    }
+
+    private void RaiseLockKeyPressed(LockKeyType keyType, bool isToggled)
```
unchihugo: I've noticed an issue with displaying lock keys status with this new method: if you press the key quickly (like a normal tap), it almost always reports the state it was previously in, and not the new state. This might be because the throttle is set too long for the lock keys, and because (like the hook version) we're listening for key presses instead of state changes.

FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
```diff
@@ -9,6 +9,10 @@
     <System:String x:Key="TextBlockText3">Display a flyout when you press a media key</System:String>
     <System:String x:Key="MediaFlyoutExcludeVolumeTitle">Exclude Volume Keys</System:String>
     <System:String x:Key="MediaFlyoutExcludeVolumeDescription">Do not display the flyout when pressing volume up/down/mute</System:String>
+    <System:String x:Key="MediaFlyoutInputSourceTitle">Input Monitor Source</System:String>
+    <System:String x:Key="MediaFlyoutInputSourceDescription">Choose how FluentFlyout detects volume changes</System:String>
+    <System:String x:Key="MediaFlyoutInputSourceOptionNAudio">NAudio</System:String>
```
unchihugo, line 14: I think we should label the hook option as "Keyboard Hook (Legacy)" so users don't necessarily think that both options are equally as good.

FluentFlyoutWPF/Services/InputMonitorService.NAudio.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// </summary>
+    private void UnbindNRenderDevice()
+    {
+        if (_nAudioRenderDevice == null)
+        {
+            return;
+        }
+
+        _nAudioRenderDevice.AudioEndpointVolume.OnVolumeNotification -= OnNVolumeNotification;
+        _nAudioRenderDevice.Dispose();
```
unchihugo: Sometimes throws user-unhandled exception "`System.Runtime.InteropServices.InvalidComObjectException: 'COM object that has been separated from its underlying RCW cannot be used'`" when disconnecting an audio device. It would probably be wise to wrap some of these methods in try/catch blocks

unchihugo (changes requested):
Hi @kitUIN, looks great so far, thank you! By migrating to NAudio events we solve the issue of flyouts only displaying from keypress events. There's some issues I've found during testing that needs addressing before we merge this update! I've written them down as comments.

I've noticed that headphones or trackpad volume manipulation successfully get detected now, though media tracking still only works from keyboard presses, so the new code unfortunately only solves the problem partially. The new method also detects volume changes from things like manually changing volume from the volume mixer, do you think we can avoid that?

Also, are you sure we still need the legacy keyboard hooks system in place? I'd like to hear your thoughts on this!

Finally, to my understanding from reading your code, there's little to no overlap with my Volume Flyout PR (https://github.com/unchihugo/FluentFlyout/pull/387) as you've only changed the hook method. I assume that you have a better understanding of how to structure the code properly, so could you have a quick look into this just to make sure? Thank you!

kitUIN:
## Second Changed

- Move constants to `NativeMethods.cs` to follow best
- Updated legacy Chinese comments to English
- Input Performance Optimization: 
  - ⛔Removed the key-press cooldown interval.
  - Replaced WPF's `System.Windows.Input.Keyboard` with the Win32 `GetKeyState` API to eliminate message loop latency.
  - Events are now triggered instantly upon detecting a state change.
- Implemented proper handling for `InvalidComObjectException` to ensure system stability during COM object lifecycle management.
- Retained the original keyboard hook mode as a temporary fallback. Given that the NAudio migration is a breaking change, this provides a safety net; it can be fully removed after the new implementation is verified across multiple production versions.

## ⚠️Known Limitation (Edge Case)

- Please note that no volume change event is triggered when the volume is already at 0% and decreased further. Consequently, the Flyout will not be displayed in this specific scenario.

## About Volume Mixer

Currently, the system only listens to System Master Volume changes. Volume adjustments at the application level (via the Volume Mixer) are ignored.

## About [Volume Flyout PR](https://github.com/unchihugo/FluentFlyout/pull/387) 

> Finally, to my understanding from reading your code, there's little to no overlap with my Volume Flyout PR (https://github.com/unchihugo/FluentFlyout/pull/387) as you've only changed the hook method. I assume that you have a better understanding of how to structure the code properly, so could you have a quick look into this just to make sure? Thank you!

I checked it.Sure that only Hook Method

unchihugo:
> Please note that no volume change event is triggered when the volume is already at 0% and decreased further. Consequently, the Flyout will not be displayed in this specific scenario.

This is a quite important edge case actually, maybe we should combine both methods instead then?

> Currently, the system only listens to System Master Volume changes. Volume adjustments at the application level (via the Volume Mixer) are ignored.

The native volume flyout and quick panel (`Win + A`) does still activate them. I wonder if there's a way to ignore those?

kitUIN:
> This is a quite important edge case actually, maybe we should combine both methods instead then?

For this edge case, using both methods might be ok

> The native volume flyout and quick panel (`Win + A`) does still activate them. I wonder if there's a way to ignore those?

Regrettably, at present I haven't found a way to ignore them.

Using NAudio to detect volume changes is a limited approach. 
Perhaps I can explore a more flexible solution.

If the sole purpose is to support the touchpad and headphones, perhaps there are better ways to do it.

unchihugo:
> If the sole purpose is to support the touchpad and headphones, perhaps there are better ways to do it.

I've just found out about [WM_APPCOMMAND](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-appcommand), most notably the WM_MEDIA and WM_VOLUME messages. If this works, leveraging this instead of NAudio will help solve all these issues we currently have with keyboard hooks and NAudio volume change events.

Would you want to develop this or should I have a go at it?

PR #497, opened 2026-03-17
[Feature] Improved Visualizer

Hikaru11747:
Improved how visualizer looks and feels, now using SDFs to draw the rounded corners (follows the win 11 rounded pill design), and optimized for better performance.

### DIFF
--- a/FluentFlyoutWPF/Classes/Visualizer.cs
+++ b/FluentFlyoutWPF/Classes/Visualizer.cs
@@ -3,11 +3,10 @@
 
 using FluentFlyout.Classes.Settings;
 using FluentFlyout.Classes.Utils;
+using Microsoft.Win32;
 using NAudio.CoreAudioApi;
 using NAudio.Dsp;
 using NAudio.Wave;
-using Microsoft.Win32;
-using System.Threading;
 using System.Windows;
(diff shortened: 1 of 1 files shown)


unchihugo (changes requested):
Hi @Hikaru11747, good performance improvements! The rounding change is a concern though — bars are now fully pill-shaped rather than the fixed radius users are already used to. Can you separate the rounding change so we can merge this performance refactor? Thanks!

Hikaru11747:
I do understand the round pill concern, will it be better if I give an option customise it by the user? A drop down to select fully squared (no radius), fixed radius and pill as styles?

unchihugo:
> I do understand the round pill concern, will it be better if I give an option customise it by the user? A drop down to select fully squared (no radius), fixed radius and pill as styles?

That's true, and I had thought about this as well. That said, I've weighed the pros and cons extensively, and I think we should keep it as the original behavior for now: we're already presenting the user with quite the amount of customization options, and a new option could contribute to more fatigue or indecisiveness (I've already noticed that most users keep settings at default other than enabling/disabling features). I'd prefer we keep the current behavior, and see whether users would like a feature like this in the future. Feel free to open an issue about this!

PR #498, opened 2026-03-18
The "Click to open Media Flyout when media is Present" function operates independently of the "Enable Media Flyout" option status

mak7im01:
The "Click to open Media Flyout when media is Present" setting no longer depends on whether the "Enable Media Flyout" is enabled.
- This is useful if you use the keyboard media keys (FN+F7, FN+F8, FN+F9, etc.). Every time you press the keyboard media keys, the "Enable Media Flyout" no longer constantly appears, and you can only control the player when you actually need to.

https://github.com/user-attachments/assets/6660a461-b16c-4ba0-8a67-625cd59898d8


mak7im01:
Hi @unchihugo , should I merge this into a commit?

unchihugo:
> Hi @unchihugo , should I merge this into a commit?

No, you're okay!

PR #499, opened 2026-03-18
[Feature] Autohide Taskbar Widget when Media is paused

zen1337th:
This PR aims to add an option to automatically hide the Taskbar Widget when Media is paused.

*"Why?"* you might ask;
> I and surely a lot of people using FluentFlyout love to customize their programs and tools they like using to match their workflow and needs. Yet I felt like this addition was somewhat missing to the Taskbar Widget, since I personally disliked seeing the last listened song with the paused symbol on the cover in my taskbar.

*Note:*
> I am currently learning software development and dedicating most of my free time to learn and improve from any mistakes. I am open for any advice towards improving my skills, quality and understanding.

### DIFF
--- a/FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
+++ b/FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
@@ -63,7 +63,7 @@ hides repeat, shuffle and player info</System:String>
     <System:String x:Key="TaskbarWidgetHideCompletelyTitle">Hide Completely When No Media Is Present</System:String>
     <System:String x:Key="TaskbarWidgetHideCompletelyDescription">Hides the entire widget when no media is available, instead of displaying a media icon</System:String>
     <System:String x:Key="TaskbarWidgetAutoHideTitle">Autohide Widget</System:String>
-    <System:String x:Key="TaskbarWidgetAutoHideDescription">Hide the widget automatically when media is paused</System:String>
+    <System:String x:Key="TaskbarWidgetAutoHideDescription">Hides the widget automatically when media is paused</System:String>
     <System:String x:Key="TaskbarWidgetControlsTitle">Widget Media Controls</System:String>
     <System:String x:Key="TaskbarWidgetControlsDescription">Show media control buttons on the widget</System:String>
     <System:String x:Key="TaskbarWidgetControlsPositionTitle">Media Controls Position</System:String>


unchihugo:
Hi @zen1337th, thanks for your interest in contributing to FluentFlyout! Unfortunately, I don't see a lot of users needing this feature currently. More features doesn't necessarily mean better — see Hick's law in UX.

Because of that, I can't merge this right now. I'll leave it open for discussion to see if other people would like something like this or if we can move the setting to the advanced settings page!

mak7im01:
@unchihugo Coincidentally, I was working on the same functionality as in this request, but with a slightly different implementation. The code is completely ready.

https://github.com/user-attachments/assets/b642bf8e-9557-4e73-8a6e-02ee6b474975

unchihugo:
Hi guys, looks like there's some demand for this feature (#554). We can go through with this - could you guys (@zen1337th and @mak7im01) discuss how to continue if you'd like to finish the implementation? Thanks!

EDIT: tagged the wrong issue

unchihugo:
Hi @zen1337th, it looks like you might've done a rebase instead of a master merge into your branch. Could you redo this? It'll be easier for me to review the code that way :)

mak7im01:
Hi @unchihugo , I'll add a hide timer function when @zen1337th  adds the basic functionality with auto-hide

zen1337th:
> Hi @zen1337th, it looks like you might've done a rebase instead of a master merge into your branch. Could you redo this? It'll be easier for me to review the code that way :)

Sorry, I missclicked on that one. Sure, I will do that ^^

zen1337th:
Hi @mak7im01, can you please further define what you mean with "adds the basic functionality with auto-hide"? If you could let me know I would be glad to make the changes you need for implementing a hide timer function for auto-hide.

Thanks in advance!

mak7im01:
Hi @zen1337th, I meant that after you add auto-hide, I'll add a timer function for it.
If you can implement it yourself, I'm fine with that. The only thing I'd like to ask/clarify is: will it look like in my video https://github.com/unchihugo/FluentFlyout/pull/499#issuecomment-4079053392?
Just to clarify, my timer is 360 seconds in 5-second increments. Setting it to zero disables the timer function.

Thanks in advance to you too.

unchihugo:
@mak7im01 if it's easier for you, you can also open PRs on @zen1337th fork to merge into their branch. This might make it easier for us to keep things managed in this PR.

zen1337th:
Hi @mak7im01 thanks for the clarification.

To answer your question: Currently, not. I have also thought about making a slider like you have in your example as well - yet a simple toggle button made more sense to me personally, as you can see in the screenshot below. The fixed value is currently 750ms until it hides automatically.
image

I have a few ideas in mind on how we can implement both of a toggle and a slider to choose the duration for the auto-hide option if you'd like to discuss those in a different PR.

I wish you guys a nice evening!

mak7im01:
@zen1337th , please specify in the description of this function (hiding) that it hides after 720 ms.
I'll wait for the update with your function to be released, and then I'll decide whether to add this timer or not.

This is all after you specified that it hides after 720 ms. I thought it would hide immediately, but it turns out it's after more than a minute.

Thank you too

unchihugo:
> @zen1337th , please specify in the description of this function (hiding) that it hides after 720 ms. I'll wait for the update with your function to be released, and then I'll decide whether to add this timer or not.
> 
> This is all after you specified that it hides after 720 ms. I thought it would hide immediately, but it turns out it's after more than a minute.
> 
> Thank you too

Just chiming in: 750ms is 0.75 seconds, not more than a minute so I think the description is okay :)

zen1337th:
> Just chiming in: 750ms is 0.75 seconds, not more than a minute so I think the description is okay :)

I think mentioning in the description that it hides after 750ms would be fine? But for you @mak7im01, here is a short video demonstrating how long it takes after the pause to hide:

https://github.com/user-attachments/assets/c40f1d94-428c-4906-a554-3144b0c48830

EDIT: If mentioning the 750ms would make sense, then I would push the change.

mak7im01:
@zen1337th , I think it's better to add it, but @unchihugo  has the final say on whether to add the time stamp (0.75 seconds) or not.

unchihugo:
We can leave it out, the visualizer does something similar without mentioning (500ms to hide when no audio is playing). Looking at the video the hide is quite abrupt, do you think that we can have some kind of fade out animation instead?

zen1337th:
> We can leave it out, the visualizer does something similar without mentioning (500ms to hide when no audio is playing). Looking at the video the hide is quite abrupt, do you think that we can have some kind of fade out animation instead?

I can see what I can do.

unchihugo:
> > We can leave it out, the visualizer does something similar without mentioning (500ms to hide when no audio is playing). Looking at the video the hide is quite abrupt, do you think that we can have some kind of fade out animation instead?
> 
> I can see what I can do.

It's alright actually, I'll review your code and merge if it's good!

FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
```diff
@@ -593,6 +596,51 @@ public void UpdateUi(string title, string artist, BitmapImage? icon, GlobalSyste
             });
             return;
         }
+        
+        if (!Dispatcher.CheckAccess())
```
unchihugo: I think we don't need this if block, just safely wrap the UI changes into Dispatcher.Invoke() functions (explained in my next comment)!

FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
```diff
@@ -593,6 +596,51 @@ public void UpdateUi(string title, string artist, BitmapImage? icon, GlobalSyste
...
+        _lastPlaybackStatus = playbackStatus;
+        
+        if ((SettingsManager.Current.TaskbarWidgetAutoHide))
+        {
+            if (playbackStatus == GlobalSystemMediaTransportControlsSessionPlaybackStatus.Playing)
+            {
+                _autoHideTimer?.Stop();
+                _autoHideTimer = null;
+                
+                Visibility = Visibility.Visible;
```
unchihugo: This should be wrapped in a Dispatcher.Invoke method so it's on the UI thread, same with line 636

``` C#
Dispatcher.Invoke(() =>
            {
                Visibility = Visibility.Visible;
            });
```

FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
zen1337th: Not sure if I understood this correctly, though removing this if block throws an InvalidOperationException with the message "The calling thread cannot access this object because a different thread owns it." as you can see in the recording below.

Keeping the if block resolved this issue in my case, which is why I added it in the first place.

https://github.com/user-attachments/assets/8c728198-6339-42b2-bf64-69367f34b0a3

unchihugo: > Not sure if I understood this correctly, though removing this if block throws an InvalidOperationException with the message "The calling thread cannot access this object because a different thread owns it." as you can see in the recording below.
> 
> Keeping the if block resolved this issue in my case, which is why I added it in the first place.
> 
>  rider64_MFOf39EWfJ.mp4

yes, this is why we wrap the UI manipulation code to the UI thread in the second comment I've added :)

PR #501, opened 2026-03-18
fix: Set DataContext after restoring settings to fix UI binding issue

### LINKED ISSUE #500: Famous bug from earlier days

### Describe the bug

everything doesnt appear even when i turned them on except for the lock keys notifications (its worth noting that when i first download everything were working fine but when i try to close it from tray icon it doesnt allow me to click my taskbar at all) when i force close it in task manager and turn it on back, checking all the settings but nothing works anymore except for the lock keys ofc

### To Reproduce

1.
2.
3.
...

_No response_

### Desktop

- OS: Windows 11 23H2
- FluentFlyout version: v2.9.0

BillionClaw:
## Problem

Issue #500: Users report that "everything doesn't appear even when turned on except for lock keys notifications." After force-closing and reopening the app, settings don't work properly.

## Root Cause

In `MainWindow.xaml.cs`, the `DataContext` was being set to `SettingsManager.Current` **before** `RestoreSettings()` was called:

1. `DataContext = SettingsManager.Current` creates a NEW default settings instance (since `_current` is null)
2. `RestoreSettings()` then replaces `_current` with the deserialized instance from the settings file
3. The UI remains bound to the old temporary instance, not the restored one
4. This causes the UI to use default values instead of the user's saved settings

## Fix

Move the `DataContext` assignment to **after** `RestoreSettings()` is called. This ensures the UI is always bound to the correct settings instance.

## Changes

- Moved `DataContext = SettingsManager.Current` from the start of the constructor to after `RestoreSettings()`

Fixes #500

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -78,7 +78,6 @@ public partial class MainWindow : MicaWindow
 
     public MainWindow()
     {
-        DataContext = SettingsManager.Current;
         WindowHelper.SetNoActivate(this); // prevents some fullscreen apps from minimizing
         InitializeComponent();
         WindowHelper.SetTopmost(this); // more prevention of fullscreen apps minimizing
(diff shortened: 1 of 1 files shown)


unchihugo:
Hi @BillionClaw, this is a great find, but I'm not sure this fixes the root cause of the issue presented because the Media Flyout is enabled by default, while users report it as not enabled (except for when you open in via Taskbar Widget).

PR #537, opened 2026-03-20
Fix taskbar widget theme detection issue

### LINKED ISSUE #524: [BUG] Taskbar widget colour not changing with app settings

### Describe the bug

With v 2.9.1 The taskbar widget text colour does not change according to the App Theme colour. The control buttons do suggesting that it supposed to change, however the text remains unreadable on my taskbar with dark wallpapers.

### To Reproduce

1. Windows Settings > Personalisation > Colours - Choose your mode: Custom, Window Mode: Dark, App Mode: Light
2. FluentFlyout Settings > System > App Theme > System Default, Dark or Light
3. Play music

Image

### Desktop

- OS: Windows 11 24H2
- FluentFlyout version: v1.9.2

PlAn-K:
Fix #524 

Fixed an issue where WindowsThemeHelper.GetCurrentWindowsTheme() failed to correctly detect the current application theme and always returned Light, by switching to MicaWPFServiceUtility.ThemeService.CurrentTheme.

There is still an issue where the Widget pause button (SongImagePlaceholder) and Taskbar Visualizer colors do not update immediately when the theme changes, requiring an function restart to reflect properly.


unchihugo:
Hi @PlAn-K, thanks for contributing! Just to make sure before we proceed, have you tested that this new method returns correct theming when the Windows theme is set to custom like the issue's description?

PlAn-K:
Yes, I have tested it and it's working great!

unchihugo:
Hmm, from the video it seems like it's still not working as expected. you can see that the taskbar widget's text is still black against a dark mode taskbar.

PlAn-K:
v2.9.2: https://github.com/user-attachments/assets/f0341ff9-0c04-4ed0-a079-ef7d5625563c
this version: https://github.com/user-attachments/assets/da35a0d4-5e80-4ce9-b33f-beed10f2ec25

In the current release, changing the app theme does not update the taskbar widget text color. This PR fixes the issue by ensuring the application theme is properly loaded and applied. (As mentioned above, this was due to the application not properly loading its theme)

The issue reported in #524 appears to be caused by the same root problem.

When the Windows OS theme is set to a custom configuration (window mode = Dark, app mode = Light), manually changing the app theme in FluentFlyout is not reflected. (Factually, since the app theme setting is not applied regardless of the Windows OS configuration, the OS settings mentioned in #524 do not seem to be relevant)

unchihugo:
Hmm, this isn't really addressing the issue - we have to detect the custom mode's theming and match the taskbar widget to it instead of following the app or Windows theme. 

I don't think it's an initialization issue and in fact, in your v2.9.2 demo it actually looks to be working correctly: following Windows system theme instead of the app's theme. What we have to do is to make it follow the correct theme when custom is enabled!

unchihugo:
Closing as another PR has handled this issue.

Issue #144, opened 2025-11-07
[FEATURE REQUEST]


jh30uk:
I would like to request you can expand the feature "Pause Other Media Automatically" to resume when done, thanks.

jh30uk:
I  would like to  bump this, it makes sense to add this as it is half done already.

jh30uk:
Bump, and fix that is does not pause sometimes and then works later.

unchihugo:
@nopeless since you developed this feature, could you look into this if you're able to and see if it's feasible? Thanks :)

nopeless:
@jh30uk you mean
if YouTube is playing, and you start playing a song in Spotify, it paused YouTube, but when you stop playing song in Spotify, it starts playing YouTube again?

this is 100% possible, but I believe it will be very annoying, but if you want, I can code the feature and you can test it out for a few days. Are you interested?

jh30uk:
That is exactly what I would like and  it can always be disabled if not liked  by others.

I listen to my music say via MusicBee but I click on a YouTube video to read about the new Intel CPU or the current WAR, ideally MusicBee pauses (it does not a lot of the time and same for all players) then when I close the YouTube video I would like MusicBee to continue without me having to do so.

Thanks for your time and if you look into this I can test it.

PR #631, opened 2026-03-31
[WIP] resume other media on pause

nopeless:
#144 

Currently, no configuration is needed. If UX is ok, supporting UI config should be introduced to this branch

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -45,6 +45,7 @@ public partial class MainWindow : MicaWindow
     private long _lastFlyoutTime = 0;
 
     public readonly WindowsMediaController.MediaManager mediaManager = new();
+    private MediaSession wasPlayingSession;
 
     // for detecting changes in settings (lazy way)
     private int _position = SettingsManager.Current.Position;
@@ -499,7 +500,10 @@ public partial class MainWindow : MicaWindow
         }
     }
 
-    private void pauseOtherMediaSessionsIfNeeded(MediaSession mediaSession)
+    /**
+     * Can either pause all or resume previous session based on setting
+     */
+    private void modifyOtherMediaSessionsIfNeeded(MediaSession mediaSession)
     {
         if (
             SettingsManager.Current.PauseOtherSessionsEnabled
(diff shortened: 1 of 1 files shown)


nopeless:
@jh30uk you can test out the build by uninstalling fluent flyout, and installing the version here (edit, wrong link)

nopeless:
actually im not sure if you have to uninstall fluent flyout, just know that it might cause problems if you have both running, be careful

nopeless:
https://github.com/nopeless/FluentFlyout/actions/runs/23824185599
latest msix per hash bccdd89

jh30uk:
It works 100% fine, and it is even more reliable then before for pausing the media.

nopeless:
@jh30uk ok, can you get back on this after you have used for a few days? to make sure there are no bugs. once you confirm, @unchihugo can make the settings ui for that

jh30uk:
It works fine and the odd glitch of not pausing the media since day one was on my daily Windows install (it is in need of a  wipe as getting glitchy) but on my test Fresh/bare min USB install it works 99.9% of the time and therefor the new resume also works.

I assume v2.10.1 will not include this function for now?

jh30uk:
Bump!

unchihugo:
Will get to this once I'm free! I'm thinking of moving this feature to Advanced Settings, since it's not necessary related to the Media Flyout (where it currently is). Are you guys fine with that?

jh30uk:
Yea one small thing, can we get a a delay as it may piss some peeps off as if you use the mouse cursor to skip forward using the  timeline in YouTube the other previous audio will resume until you click the mouse again (so if you are clicking the timeline multiple times to find a certain spot its a bit of a PIA).

It does not do this if you use the cursor keys to skip forward on the timeline.

A setting to let us choose between 0-5secs would fix this IMO.

nopeless:
> Will get to this once I'm free! I'm thinking of moving this feature to Advanced Settings, since it's not necessary related to the Media Flyout (where it currently is). Are you guys fine with that?

I don't see myself using the feature, but it seems appropriate



> Yea one small thing, can we get a a delay as it may piss some peeps off as if you use the mouse cursor to skip forward using the timeline in YouTube the other previous audio will resume until you click the mouse again (so if you clicking the timeline many times to find a certain spot its a bit of a PIA).
> 
> It does not do this if you use the cursor keys to skip forward on the timeline.
> 
> A setting to let us choose between 0-5secs would fix this IMO.

@jh30uk nice catch. this is why you don't really roll out features fast, like, I never could have thought of that happening. I'll look into this for a bit

nopeless:
I looked into the api and there isn't an explicit "is scrubbing", but there is a "change position" event, meaning that we can kinda guess what the user is doing by checking their scrubbing state. Does this seem like a good idea?

jh30uk:
Sure and I will gladly test it for you.

nopeless:
@jh30uk @unchihugo while implementing the feature, I found a bug?, the resume feature just completely messes up user experience. I need a resolution for this #675

PR #652, opened 2026-04-04
Add App Filtering Page and Session Validation (ref #633)

### LINKED ISSUE #81: [FEATURE REQUEST] Application Whitelist/Blacklist or application selection option
**Purpose:**  
I just recently started using this application and it has been working great, however I notice that all media shows up within the flyout, personally I only want the flyout to be for my music player, and not for videos in my browser.

**Description:**  
The best way I can see this working is having the setting menu allow you to chose what application/service should show up. For example I am using foobar2000, so having a way to filter the flyouts, especially the upcoming, would be amazing if we could chose the service/application. I understand that since foobar2000 is not your traditional music player, having the option to instead exclude specific applications/services would be a good compromise. 

**Additional context:**  
Overall I think this application is real nice, and I am excited to continue to use it. If you have any questions including information regarding foobar2000 (Such as a Component Report), I would be happy to provide anything you need. 

I am using Windows 11 24h2, with the latest version of Foobar2000 (Noteworthy components: `foo_cad_plus` and `foo_beefweb`)

Side Feature: Having a widget for windows 11 would be a nice feature, as I haven't found one that works properly, but this would definitely be a more complicated request.

### LINKED ISSUE #617: [FEATURE REQUEST] Blacklist for sites/app pause
### Description

Would it be possible to add a blacklist for websites that we don't want to pause the music?

For example, if I'm listening to Spotify and i open a YouTube video, it's fine for the music to pause, but if I open Panzer Quest (a game on Chrome with sounds), I don't want the music to stop.

### Additional context

_No response_

### LINKED ISSUE #633: [FEATURE REQUEST] Integration with music app
### Description

It would be nice if FluentFlyout has an integration or link to a music app (example: Spotify or YouTube Music) so I can control it from the taskbar without having to open my music app. I know that I can currently control a music app with FluentFlyout,
but the problem is when you start playing other multimedia content FluentFlyout changes and stops showing the content of the music app

### Additional context

_No response_

galagyy:
## Summary
This PR introduces an app filtering feature. I had seen a feature request, specifically #633, which discusses an idea for a feature similar to this.

This allows the user to create a custom "Allow list" and "Block list" where users may pick which apps they want affecting the flyout and taskbar player.

## Motivation
As mentioned above, feature request #633 introduced an idea of a feature similar to this. This request has more features they want added so I do not suggest closing it yet.

## Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Other

## What Changed
- App Filtering Page: Created an app filtering page where users can define custom rules on which apps to consider for the flyout & taskbar. The page has a dropdown of open processes and an addition section to add apps by program executable name.
- Home & Settings Navigation: Added the aforementioned App Filtering page to the dashboard and added it to the left-side navigation.
- Media Logic Changes: Added a new function, `IsSessionAllowed()`, to validate the current session. This function has replaced the previous `GetFocusedSession()`.
- Localization: Added the text and such for the new page to the localization file.

## Additional Information
Originally I had the filter list housed under the taskbar section; I have moved it to its own section due to how it affects both the taskbar and flyout.

NOTE: I have also gone ahead and added the copyright disclaimer on top of the files added, but I am unsure if I was supposed to do that.

## Checklist
- [x] AI tools were used (Used to verify changes are accurate & fix formatting issues).


galagyy:
A quick comment about the formatting, I come from a Java/C++ background so I had originally formatted it with braces on the same line. I changed it in one of the commits, but if there is still an issue with it let me know and I will try my best to fix it.

Thanks!

galagyy:
Found a small bug regarding pause/play, I will try fixing it today.

galagyy:
Fixed the bug above, I changed the old implementation that relied on manually sending a keyboard input.

FluentFlyoutWPF/MainWindow.xaml.cs
```diff
@@ -515,20 +571,26 @@ private void CurrentSession_OnPlaybackStateChanged(MediaSession mediaSession, Gl
 #endif     
         pauseOtherMediaSessionsIfNeeded(mediaSession);
 
-        var focusedSession = mediaManager.GetFocusedSession();
-        if (focusedSession == null)
+        var taskbarSession = GetActiveMediaSession();
+        if (taskbarSession == null)
         {
             taskbarWindow?.UpdateUi("-", "-", null, GlobalSystemMediaTransportControlsSessionPlaybackStatus.Closed);
-            return;
```
unchihugo, line 553: I believe the early return here is justified unless I'm mistaken, and adding the else clause isn't necessary either. I think we should bring the structure back to a similar state of what it was before (early return if null, no else needed).

What happens now when focusedSession == null:

taskbarSession = GetActiveMediaSession(); // first check, null
update taskbar ui, but don't return
focusedSession = GetActiveMediaSession(); // a second time, null again
if (focusedSession == null) return; // returns after calling same method again

Additionally, variable taskbarSession is the same as focusedSession. We should keep the focusedSession naming!

FluentFlyoutWPF/MainWindow.xaml.cs
```diff
@@ -576,7 +639,23 @@ private void MediaManager_OnAnyMediaPropertyChanged(MediaSession mediaSession, G
 
         var thumbnail = BitmapHelper.GetThumbnail(songInfo.Thumbnail);
         BitmapHelper.GetDominantColors(1);
-        taskbarWindow?.UpdateUi(songInfo.Title, songInfo.Artist, thumbnail, playbackInfo.PlaybackStatus, playbackInfo.Controls);
```
unchihugo, line 610: If I'm not mistaken, this change is unnecessary right? Correct me if I'm wrong!

FluentFlyoutWPF/Pages/AppFilteringPage.xaml.cs
```diff
@@ -0,0 +1,141 @@
+// Copyright � 2024-2026 The FluentFlyout Authors
```
unchihugo: For some reason, the copyright symbol is a question mark here.

unchihugo (changes requested):
Hi @galagyy, solid implementation so far! Just a few things I've noticed throughout that we should adjust before merging. I've noted it in the comments. 

I'd like to discuss the implementation of the filter lists: 
Currently, names in the filter lists aren't sanitized - we should have this feature utilize our [MediaPlayerData class](https://github.com/unchihugo/FluentFlyout/blob/master/FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs). Methods in here can find the sanitized title + icon for apps/media players, especially with the changes I've added myself in the Volume Flyout branch [here](https://github.com/unchihugo/FluentFlyout/blob/81b30eb25db3fb0f8f2db2d64ad16bea12fddb19/FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs). 

Could you adapt the current code to use the GetMediaPlayerData() method in MediaPlayerData instead of ExtractAppName()? That would be great!

Once the current issues/questions are addressed, we can continue with the merge :)

galagyy:
Hello, thank you for the prompt review!

I have addressed all the issues you pointed out earlier:
- Reverted back to the prior naming scheme & format
- Reverted to using pre-existing sanitization & function
- Re-added the early `return` statement
- Fixed the malformed copyright symbol
- Fixed the redundant call to retrieve album image, name, etc.

I also had noticed that all the other tabs have an image describing their function with this specifically standing out as one without; would you like me to try to find an image suitable for it?

Thank you!

FluentFlyoutWPF/MainWindow.xaml.cs
galagyy: Yes, you are correct! I have gone ahead and fixed it.

FluentFlyoutWPF/Pages/AppFilteringPage.xaml.cs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private void PopulateComboBox(ComboBox comboBox) 
+    {
+        var mainWindow = (MainWindow)Application.Current.MainWindow;
+
+        if (mainWindow?.mediaManager != null) 
+        {
+            var apps = mainWindow.mediaManager.CurrentMediaSessions.Values
+                .Select(s => MediaPlayerData.getMediaPlayerData(s.Id).Item1)
```
unchihugo: Could you explain why you have to specify `Item1` here?

FluentFlyoutWPF/MainWindow.xaml.cs
```diff
@@ -229,6 +229,62 @@ private async Task CheckForUpdatesOnStartupAsync()
...
 
+    public bool IsSessionAllowed(MediaSession? session)
+    {
+        if (session == null) return false;
+        if (!SettingsManager.Current.AppFilteringEnabled) return true;
+
+        string appId = session.Id ?? string.Empty;
+        string appName = MediaPlayerData.getMediaPlayerData(appId).Item1 ?? appId;
+
+        if (SettingsManager.Current.BlockedApps != null && SettingsManager.Current.BlockedApps.Any(b => appName.Contains(b, StringComparison.OrdinalIgnoreCase) || appId.Contains(b, StringComparison.OrdinalIgnoreCase)))
```
unchihugo: Could you also explain why we also have to check with the sanitized player name?

unchihugo (changes requested):
Hi, thanks for the quick refactor, though could you answer my questions I've noted in the code review? Just asking since I'm a bit busy and can't load up the build right now :)

> I also had noticed that all the other tabs have an image describing their function with this specifically standing out as one without; would you like me to try to find an image suitable for it?

I was thinking of having this be a subpage from the System page instead (kind of like the Advanced Settings button there). Therefore we wouldn't need an image for it either currently!

PS: have you tested whether the manual adding works?

FluentFlyoutWPF/Pages/AppFilteringPage.xaml.cs
galagyy: I chose to specifically pick `Item1` because the `getMediaPlayerData()` function returns both `(string, ImageSource)` and I specifically need only the title of the application.

FluentFlyoutWPF/MainWindow.xaml.cs
galagyy: I added it more so as a fallback in case a more advanced user may want to block an app with its specific process name or its ID (e.g. chrome.exe over chrome); I also kept a check against the `appId` to ensure the filter will always catch what the user enters. I could keep it one system if you'd like though.

galagyy:
Hello, thanks for the response!

I will try moving it to the Systems page as you specified earlier. Regarding manual adding, it has worked with the apps I've tried (e.g. Chrome, Edge, Firefox, etc.) but I haven't come across an app that wouldn't be displayed on the default dropdown yet; I will try to test that case soon.

I will have the changes done hopefully by the end of today; I will try adding pictures or a video if you are unable to build it currently.

unchihugo:
Okay, that makes sense - will wait for updates!

galagyy:
Currently I have these for the system, I will post further images/videos in some time.

I've removed the tab for App Filtering in the left-side hamburger menu and placed it here instead.
image

The UI for the actual App Filtering page looks a little something like this:
image

I was able to confirm that manually adding an app works for all apps that I had tested from. I am working on removing it from the main page and tidying up the code a bit.

unchihugo:
That's good! 

I'm thinking about the user experience with the current UI - it took me a bit to figure out what was happening myself, so I think we should definitely improve UX before we ship this to average users. 

We can brainstorm this after you're ready. A lot of users on GitHub called this feature something like whitelist/blacklist instead of filtering.

galagyy:
Hello! Yes, the UX needs to be improved, if you have any ideas I am open to them as I am not very good at UX myself.

I'm about to commit the changes where the button is removed from the hamburger menu and main menu; it should hopefully be a matter of naming and UX now.

galagyy:
I went ahead and added a few changes for the manual addition, namely better comparison against known sources to prevent two rules for the same app.

I also went ahead and added documentation to all the methods for the `AppFilteringPage.xaml.cs` for future maintainability.

I believe the only hurdle remaining as you said should be the UI now!

galagyy:
I've tried making it more friendly, using the whitelist and blacklist language you had pointed out earlier; I am still unsure if it is intuitive to the average user though.

image

FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
```diff
@@ -171,6 +171,19 @@ Note: As an open-source project, these features are also available for free via
     <System:String x:Key="TrayIconSectionTitle">Tray Icon</System:String>
     <System:String x:Key="UpdatesSectionTitle">Updates</System:String>
     <System:String x:Key="SystemConfigure">Configure</System:String>
+    <System:String x:Key="AppFilteringTitle">Whitelist / Blacklist</System:String>
+    <System:String x:Key="AppFilteringRulesTitle">Whitelist and Blacklist</System:String>
```
unchihugo: Could you change "AppFilteringTitle" to "Whitelist & Blacklist" and remove "AppFilteringRulesTitle"? The page can use "AppFilteringTitle" instead!

unchihugo (changes requested):
Look great @galagyy! I think the renaming definitely makes things a bit more clear. We're almost ready to add this feature in, but I've noticed that if you click a media player in the list it adds a colored background (indicating select), are you able to remove that? Thank you!

By the way, I'd love to have icons displayed next to the name in the future; if you'd like contributing more, do you think you could handle that as well?

galagyy:
I'll work on the changes when I get back today, they should be ready by the end of today. I will try adding the icons as that seems like a great addition.

darklinkpower:
Hi, first of all, thank you for working on this feature, it looks great. Hope you don't mind me chiming in with an outsider perspective, but I think there are a few potential issues with the proposed implementation.

I think having both a whitelist and blacklist in their current form introduces some ambiguity and UX friction.

Right now there are effectively two concurrent filtering models:

- Blacklist -> block these apps (default allow)
- Whitelist -> only allow these apps (default deny)


The issue is that once the whitelist has any entries, it implicitly overrides the blacklist, making it effectively unused. This creates:

- Ambiguity (which list takes precedence, if any? The same program can also be in both lists?)
- Hidden behavior (mode switches based on whether whitelist is empty)
- Potential confusion for users trying to understand why something is or isn't showing
- The user needs to wonder and choose: Should I use the whitelist, the blacklist, or maybe both?

I'm wondering if it might be clearer to make this a single-mode selection feature in a simple dropdown/ComboBox, for example:

a) Blacklist mode: Allow all except blocked apps
b) Whitelist mode: Only allow selected apps

Then show only the relevant list based on the selected mode.

This would make the behavior explicit, avoid overlap, and simplify the mental strain for users, resulting in a clearer UX. I'd suggest making "Blacklist" (allow-all except blocked) the default, since it preserves current behavior. If "Whitelist mode" is selected, it should behave consistently regardless of whether the list is empty, meaning no implicit mode or logic-behavior switching based on list contents (Is the list empty or not?). In my opinion avoiding that kind of hidden logic would make the feature more predictable and easier to understand.

Apologies if this feedback is unnecessary or wrong, especially when the developer is okay with the current approach. Just wanted to share my perspective in case it's helpful.

unchihugo:
Hi @darklinkpower, you're making great points against the current UX. I was also brainstorming about the dropdown -> whitelist / blacklist feature instead of presenting both at once, as I had some difficulties understanding the UI as well. Your feedback definitely clarify the reasons to change implementation, and is incredibly helpful to us. 

That said, @galagyy if you're able to, could you change the implementation to a dropdown with blacklist and whitelist options separated? I can take over the UI if you'd like!

galagyy:
Hello, @darklinkpower made some great points surrounding the UI, the dropdown feature seems like a much better fit than what I had created.

Regarding the UI, I intend to push the icon change and try to implement the feature in a little bit of time; if you could take over the UI after that would be great!

galagyy:
I believe I should have the app icon implemented properly; I added in retrieving an icon from disk if it is not in cache as some apps may not be open and still may be on either the whitelist or blacklist.

I tried my best with changing the UI to meet the ideas and address the issues you both had outlined earlier. I am still new to C# and .NET though so I cannot say it works 100%, but I believe it should be fine. The format feels as if it's missing something but I'm unsure myself what that could be.

If you could look at the UI/take over the UI now that would be great!

unchihugo:
@galagyy I'll have a look at it once I'm free! By the way, are you able to exclusively use the MediaPlayerData method? No worries if it doesn't return an icon for now. We have some MediaPlayerData icon utilities in the Volume Flyout branch, and adding more methods might make things too complicated or add some redundancy. We can also work on icons in another branch/PR if you prefer to get this feature merged first.

galagyy:
Hello! Yes, I could do that; do you want me to add the disk retrieval to the `MediaPlayerData` method or remove the idea completely for now? 

I could also go about using the old method but having the icon only when the blacklisted/whitelisted application is open doesn't sound like the best idea.

unchihugo:
> Hello! Yes, I could do that; do you want me to add the disk retrieval to the `MediaPlayerData` method or remove the idea completely for now?

You can remove it entirely for now. The thing is, in the Volume Flyout branch there's new methods to find icons and names more accurately automatically, so I think we should see how that works out first once both branches are merged!

galagyy:
Alright, I've gone ahead and removed the icons for now as you suggested! I'll await the Volume Flyout's implementation later.

galagyy:
> Could you change "AppFilteringTitle" to "Whitelist & Blacklist" and remove "AppFilteringRulesTitle"? The page can use "AppFilteringTitle" instead!

I somehow glossed over this, I have gone through the code and removed the redundant one.

galagyy:
> That said, @galagyy if you're able to, could you change the implementation to a dropdown with blacklist and whitelist options separated? I can take over the UI if you'd like!

If you would like, @unchihugo, I could try implementing the UI changes/ideas you had yourself; I spent some time looking into UI and I feel I can try implementing those changes.

unchihugo:
I've been quite busy lately, sorry! I will definitely get around to looking into this but you can have a go at it if you wish!

galagyy:
All good! I'll try a few things in a bit of time.

galagyy:
I'm going to try to fix the merge conflicts this weekend and try the UI changes; I had a few planned but did not get around to adding it.

unchihugo:
Sounds good! I've also found a similar feature in MS PowerToys - perhaps you could find inspiration or design the elements similarly to this? It doesn't look confusing and gets the point across really quickly
image

galagyy:
Good idea! I'll try having that UI done by Monday as well as the merge conflicts!

galagyy:
I just fixed the merge conflict, I will work on the UI shortly.

galagyy:
Hello,

Here are the UI changes for the app filtering page; I tried keeping it similar to the PowerToys reference image @unchihugo gave while keeping the dropdown and custom add option.

I've also updated the dictionary definitions to be more clear to the user with simpler language.

## Link to the page
image

## Dropdown to pick between whitelist and blacklist
image

## Whitelist option
> This is the entire page with the whitelist option selected.
image

## Blacklist option
> This is the entire page with the blacklist option selected.
image

unchihugo:
Thank you @galagyy, looks great! I'll test it out soon :)

galagyy:
I asked a few people for feedback with the UI and I've gone ahead and added these slight changes in response to what I heard:
- Added an icon to "Whitelist" and "Blacklist" options in the main dropdown
- Added a hint to app dropdown which prompts users to select an app to add to the filter (a little bit of a workaround code-wise, but it should work properly)
- Changed "Whitelist" and "Blacklist" options to be centered when opening the page
- Fixed a small issue with selecting the "Whitelist" or "Blacklist" option in the main dropdown

Nothing massive changed logic-wise, just mainly UI changes. With respect to the dropdown alignment change, I can change it back if you feel it to be a bit off.

### Example:
image

unchihugo (changes requested):
Just checked out the UI improvements, and it's definitely on the right path! I would suggest a few changes before it's ready:
1. The blacklist/whitelist dropdown should be aligned to the left instead of center - I definitely like how it looks aligned to the left before!
2. The Allowed Apps element should be collapsible (and opened by default) if possible, though you can keep the functionality if it's too time consuming to implement.
3. The Allowed Apps element also looks to be having more rounded corners, let's keep it the default Fluent UI style.

Finally, not sure why but during testing some of my media players' names were different/incorrect compared to the current release version: Microsoft Edge for example turned into Edge Webview2. Looking at your code I don't see many modifications to how names are found other than the removed lines at `MediaPlayerData.cs`. Could you look into this?

galagyy:
Hello,

Thank you for the comment! I will attempt to fix those issues today.

darklinkpower:
Hey, sorry to intrude again. The PR is looking good but if you'd allow me to make one final suggestion, I'm wondering if there should be a Label next to the mode dropdown (`Whitelist (Only allow selected apps)`). In my opinion to provide a clear UX it should say exactly what it is for rather than have it just there with no label, and it's something already done everywhere else in the app, for example:

image

The "App Language" case is very good in my opinion because it also adds a smal subtext with an explanation. I think a good Label for this element could be "Execution Mode"

galagyy:
That's a great point, I like the "App Language" case you brought up; I'll try changing it to that in some time after I address the improper app name issue.

On that case, do you have an idea of what the explanation could be?

galagyy:
With the image above, I've changed it to what @darklinkpower suggested with a short message explaining it, although I can change it to a more comprehensive one if everyone believes that would be better.

image

unchihugo:
We could try "Filtering Mode", since I think it sounds clearer than Execution Mode. As for the subtitle: the current one sounds a bit technical and I'm not sure if the average user will understand what it means. Any other ideas for that?

Btw, when you add strings to the dictionary and we merge the PR to master, all translators will work with those merged strings, that's why I believe it's important to get the strings right before merging :)

galagyy:
> Btw, when you add strings to the dictionary and we merge the PR to master, all translators will work with those merged strings, that's why I believe it's important to get the strings right before merging :)

That's a good point with the strings, I have a few ideas of what we could change it to in order to make it more user friendly. 

I agree with changing the title to "Filter Mode," as that is more representative of what it's actually doing. For the description, I was thinking we could change it to something simple such as "Choose which filter list is active," although we could make it more informative if everyone feels that's better.

I was also thinking of changing the dropdown options from Whitelist/Blacklist to Allow List/Block List as other applications (Google's UI, Apple's UI, etc.) have started using this to be more user friendly.

> Finally, not sure why but during testing some of my media players' names were different/incorrect compared to the current release version: Microsoft Edge for example turned into Edge Webview2. Looking at your code I don't see many modifications to how names are found other than the removed lines at `MediaPlayerData.cs`. Could you look into this?
@unchihugo 

With regards to this, I believe I have fixed the issue as it no longer replicates on my own machine (I removed a piece checking for background workers when I shouldn't have); do you have the names of the other apps that caused this issue?

Thanks,

darklinkpower:
> We could try "Filtering Mode", since I think it sounds clearer than Execution Mode. As for the subtitle: the current one sounds a bit technical and I'm not sure if the average user will understand what it means. Any other ideas for that?

I think "Filtering Mode" sounds great. Regarding the subtitle, how about:

> Choose whether listed apps are allowed or blocked

In my opinion it pairs well with the top "Enable Filtering" toggle, although I'd change this one to simply "Control which applications can display media" to explain the feature at a high level, whereas the dropdown goes into more specifics.

image

galagyy:
Thanks for the suggestions, this is what it's looking like right now:

image

I'm working on making it (the Allowed/Excluded Apps sections) a dropdown as suggested before.

galagyy:
> Just checked out the UI improvements, and it's definitely on the right path! I would suggest a few changes before it's ready:
> 
> 1. The blacklist/whitelist dropdown should be aligned to the left instead of center - I definitely like how it looks aligned to the left before!
> 2. The Allowed Apps element should be collapsible (and opened by default) if possible, though you can keep the functionality if it's too time consuming to implement.
> 3. The Allowed Apps element also looks to be having more rounded corners, let's keep it the default Fluent UI style.
> 
> Finally, not sure why but during testing some of my media players' names were different/incorrect compared to the current release version: Microsoft Edge for example turned into Edge Webview2. Looking at your code I don't see many modifications to how names are found other than the removed lines at `MediaPlayerData.cs`. Could you look into this?

@unchihugo All the changes you had pointed out should be complete now. I have tested it with a fair share of apps, but I will try testing/looking for more edge cases.

image

unchihugo (changes requested):
Hi @galagyy, I'm ready to merge after these changes:
1. Could you move the `MediaPlayerData.cs` changes out of this current branch? They don’t seem relevant to the PR (unless I'm mistaken, I'd like to hear your thoughts), so separating them would be helpful.
2. Also, I think it’s better to keep blacklist as the default rather than switching to whitelisting :)

galagyy:
Hello,

Thanks for the prompt reply! I've gone ahead and made the blacklist the default mode as pointed out.

Regarding the changes to `MediaPlayerData.cs`, I chose to include them in the PR because they are imperative to how candidate apps are found, which is needed for the string-matching approach the App Filtering implementation has.

I noticed that certain apps would fail to filter with the old code. Since some apps handle audio through background workers, the old check against `IntPtr.Zero` would occasionally skip them entirely. Furthermore, once we allow headless apps through, we have to sort them properly and fallback to `ProductName`; if we do not, the code grabs the first headless worker it finds and populates the filter list with unfriendly names (like "Microsoft Edge WebView2" as you had seen earlier), which breaks the user's whitelist/blacklist text matching.

As a result of the above I feel the changes are needed specifically for the app filtering, but I could try testing again with the old implementation as I originally tested a few commits ago; I do doubt the code would work though with the caveats pointed above.

unchihugo:
Thanks for the follow-up! I see the issue now. I'll have a look into this.

galagyy:
I'm trying to fix the dotnet format issue but I'm unable to figure out why it's failing; I looked at the EOF issue the formatter had mentioned but am unable to find it on my end.

unchihugo:
> I'm trying to fix the dotnet format issue but I'm unable to figure out why it's failing; I looked at the EOF issue the formatter had mentioned but am unable to find it on my end.

I just pushed commit https://github.com/unchihugo/FluentFlyout/commit/c1ec46caec638d844b782d9b7855850b228b36a2 to master to fix the issue. You can pull the changes to your branch and it should fix the check!

galagyy:
I've gone ahead and pulled the changes, but I still get this particular error: `FINALNEWLINE: Fix final newline. Delete 2 characters.`

I've checked the newline and it appears to be fine, I'm not too sure what it's asking me to resolve.

unchihugo (changes requested):
Hi @galagyy, I just tested the changes and it's great - we're nearing the finish now and I'm excited to get this feature in the app with you. I found a breaking issue though: 

The new `MediaPlayerData.cs` is incredibly CPU resource intensive (I believe it's searching through way too many processes now), resulting in ~3-5 second freezes when the `GetAndCacheMediaPlayerData` method tries to find a new media player. On my end, when I reverted the changes to that file back to master, it works flawlessly. Perhaps it would be better for us to revert the changes to this file and start a new PR for background process tracking.

Another thing I've noticed is that `GetActiveMediaSession()` in `MainWindow.xaml.cs` searches for the focused session session first, then falls back to the first playing session. Previously it would not check if the status was playing: this is a change not related to the PR, and I'm thinking we should simply return validSessions.FirstOrDefault() instead of check whether the session is playing like how it used to work, and open a now PR for this as well.

I'm sorry I'm dragging this PR, but I definitely believe that after these two changes we can safely merge the PR to master!

EDIT: I'll handle the formatting later, it's okay to leave it!

galagyy:
Hello! No worries for dragging out the PR request, in the end it's better for everyone! I've learned a lot about C# from this PR.

With regards to the issues, I will try reverting them and making sure it works properly with the code.

darklinkpower:
> The new `MediaPlayerData.cs` is incredibly CPU resource intensive (I believe it's searching through way too many processes now), resulting in ~3-5 second freezes when the `GetAndCacheMediaPlayerData` method tries to find a new media player. On my end, when I reverted the changes to that file back to master, it works flawlessly. Perhaps it would be better for us to revert the changes to this file and start a new PR for background process tracking.

I'm probably missing something but I think it's an easy fix. Right now the issues I see in the loop are:
- References to `mainModule` for each item in the loop that are really expensive
- It analyzes all processes to only use one of them in the end.

To fix this it may be enough to leave the loop early when a process with the max possible priority is found.

image

---

Here's how I'd modify the method, apologies if formatting is incorrect or if there's something wrong, I wrote it with a text editor and might have missed something:

```csharp
    private sealed class ProcessMatch
    {
        public required Process Process { get; init; }
        public required string Path { get; init; }
        public required int Priority { get; init; }
    }
    
    public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlayerId)
    {
        if (mediaPlayerCache.TryGetValue(mediaPlayerId, out var cachedInfo)
            || mediaPlayerIdVariants.TryGetValue(mediaPlayerId, out var variantKey)
            && mediaPlayerCache.TryGetValue(variantKey, out cachedInfo))
        {
            return (cachedInfo.Title, cachedInfo.Icon);
        }

        string mediaTitle = mediaPlayerId;
        ImageSource? mediaIcon = null;

        // get sanitized media title name
        string[] mediaSessionIdVariants = mediaPlayerId.Split('.');

        // remove common non-informative substrings
        var variants = mediaSessionIdVariants.Select(variant =>
            variant.Replace("com", "", StringComparison.OrdinalIgnoreCase)
                   .Replace("github", "", StringComparison.OrdinalIgnoreCase)
                   .Replace("exe", "", StringComparison.OrdinalIgnoreCase)
                   .Trim()
        ).Where(variant => !string.IsNullOrWhiteSpace(variant)).ToList();

        // add original id to the end of the array to ensure at least one variant
        variants.Add(mediaPlayerId);

        Process[] processes;

        // use cache to avoid frequent process enumeration
        if (cachedProcesses == null || (DateTime.Now - lastCacheTime).TotalSeconds > CACHE_DURATION_SECONDS)
        {
            cachedProcesses = Process.GetProcesses();
            lastCacheTime = DateTime.Now;
        }

        ProcessMatch? bestProcessMatchInfo = null;

        foreach (var p in cachedProcesses)
        {
            try
            {
                // we no longer instantly filter out processes without a main window handle, 
                // to support background media processes (chrome, edge, etc. may spawn background
                // workers).
                var module = p.MainModule;
                if (module == null)
                {
                    continue;
                }

                string path = module.FileName;
                bool matches = false;

                foreach (var variant in variants)
                {
                    if (path.Contains(variant, StringComparison.OrdinalIgnoreCase))
                    {
                        matches = true;
                        break;
                    }
                }

                if (!matches)
                {
                    continue;
                }

                // prioritize processes that have a main window, as they represent the primary app (e.g., Microsoft Edge) rather than a background worker
                int priority = p.MainWindowHandle != IntPtr.Zero ? 1 : 0;

                if (bestProcessMatchInfo == null || priority > bestProcessMatchInfo.Priority)
                {
                    bestProcessMatchInfo = new ProcessMatch
                    {
                        Process = p,
                        Path = path,
                        Priority = priority
                    };

                    // Break early if maximum priority was obtained
                    if (priority == 1)
                    {
                        break;
                    }
                }
            }
            catch (Win32Exception)
            {
            }
            catch (InvalidOperationException)
            {
            }
        }

        if (bestProcessMatchInfo == null)
        {
            return (mediaTitle, mediaIcon);
        }
        
        try
        {
            // prioritize ProductName for the user-facing app name, but fall back to `FileDescription` or `MainWindowTitle` if `ProductName` is not available
            var versionInfo = bestProcessMatchInfo.Process.MainModule?.FileVersionInfo;

            mediaTitle =
                !string.IsNullOrWhiteSpace(versionInfo?.ProductName)
                    ? versionInfo.ProductName
                : !string.IsNullOrWhiteSpace(versionInfo?.FileDescription)
                    ? versionInfo.FileDescription
                : bestProcessMatchInfo.Process.MainWindowTitle;
        }
        catch
        {
        }

        // check cache again because we have the sanitized title
        if (mediaPlayerCache.TryGetValue(mediaTitle, out cachedInfo))
        {
            // map the original id to the sanitized title for future lookups
            mediaPlayerIdVariants[mediaPlayerId] = mediaTitle;
            return (cachedInfo.Title, cachedInfo.Icon);
        }

        mediaIcon = GetIconFromPath(bestProcessMatchInfo.Path);

        mediaPlayerCache[mediaPlayerId] = new CachedMediaPlayerInfo
        {
            Title = mediaTitle,
            Icon = mediaIcon,
            ProcessId = bestProcessMatchInfo.Process.Id
        };

        return (mediaTitle, mediaIcon);
    }
```

The idea is to break early from the loop if a match is found and move expensive calls outside the loop, so they are only processed for a single one of them. This also removes a few LINQ processing calls although it's not like they would make any realistic difference.

galagyy:
Hello! Sorry for the short hiatus, I'll work on this in a little bit; something had came up in the past week.

unchihugo:
Hi @galagyy, just wondering, will you be able to finish this soon? I'd love for us to get this into the app!

galagyy:
Hello! I apologize for the long wait, I have the changes done and will commit them sooner today.

galagyy:
I'm currently removing the feature that dictated which session to show as you had pointed out earlier; I added it initially since I thought it was a good feature to tie in with this but I will move it to a separate PR in the future, that makes more sense!

unchihugo:
Thanks for the fixes! I think the MediaPlayerData file is quite messy at this stage, and we should just revert it to what it is on master. Once that's done, we're ready to merge!

galagyy:
Hello, thank you for the response! I will go ahead and do that.

unchihugo (approved):

unchihugo:
Perfect, works amazingly now! Thank you so much for the contribution :)

galagyy:
@unchihugo Thank you! Also, regarding the scrolling text feature, should I wait for the app filtering PR to be merged into master and branch off there, or should I just branch directly off the app filtering branch for now?

unchihugo:
For that feature, you should update your own fork to upstream, then merge your scrolling text branch with master. If merging adds new files or changes, try undoing and removing them from the branch, leaving only the text scrolling features in there. Then you should be ready to create a new PR!

EDIT: for future reference, to avoid issues like these you should always create a new branch from master/main instead of a specific unmerged branch, unless you absolutely need to :)

PR #852, opened 2026-06-05
Add locale-aware GlobalFontFamily and enable dynamic FontFamily

lindwurm:
## Summary

Switched to changing the font family dynamically to use correct one.

|before|after|
|---|---|
|||

## Motivation

The language‑specific font priority isn't working in both the Media Flyout and the Taskbar Widget ("Up Next" isn't affected). As a result, characters aren't rendered using the intended fonts (In the following example, selected font is not suitable for displaying Japanese).

> 
> 
> title `おしえてアンドロメダ` in "Up Next" is using `Yu Gothic UI` (Japanese font), but Taskbar Widget isn't.

For Japanese users, being forced into an inappropriate fallback font significantly degrades readability and user experience — much like replacing all English text with Comic Sans🙃

```cs
  // dictionary of font families for specific languages, priorities are switched around
  private static readonly Dictionary<string, string> _languageFontFamilies = new()
  {
      { "default", "Segoe UI Variable, Microsoft YaHei UI, Yu Gothic UI, Malgun Gothic" }, // default support for multiple languages
      //{ "zh-CN", "Segoe UI Variable, Microsoft YaHei UI, Yu Gothic UI, Malgun Gothic" }, // same as default
      { "zh-TW", "Segoe UI Variable, Microsoft JhengHei UI, Yu Gothic UI, Malgun Gothic" },
      { "ja", "Segoe UI Variable, Yu Gothic UI, Microsoft YaHei UI, Malgun Gothic" },
      { "ko", "Segoe UI Variable, Malgun Gothic, Microsoft YaHei UI, Yu Gothic UI" },
  };
```

The logic was in place (eg. #317 ), but it wasn’t actually taking effect in all relevant parts of the UI.

## Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Other

## What Changed

- Added locale-aware global font selection in LocalizationManager.cs
  - Applied dynamic `GlobalFontFamily` via `Application.Current.Resources` so UI updates at runtime
  - Kept existing localization loading logic while enabling dynamic font switching based on current culture
- Ensured `TextBlock` and `Control` styles in App.xaml use `DynamicResource GlobalFontFamily` for consistent font application
  - Excluded MicaWPF SymbolIcon from `GlobalFontFamily` so icon rendering is not broken

## Checklist
- [ ] Formatting and naming are consistent with the project.

### DIFF
--- a/FluentFlyoutWPF/App.xaml
+++ b/FluentFlyoutWPF/App.xaml
@@ -6,6 +6,7 @@
              StartupUri="MainWindow.xaml"
              xmlns:ui="http://schemas.lepo.co/wpfui/2022/xaml"
              xmlns:mica="clr-namespace:MicaWPF.Styles;assembly=MicaWPF"
+             xmlns:controls="clr-namespace:MicaWPF.Controls;assembly=MicaWPF"
              xmlns:utils="clr-namespace:FluentFlyoutWPF.Classes.Utils"
              ShutdownMode="OnExplicitShutdown">
     <Application.Resources>
@@ -21,7 +22,23 @@
                 <ResourceDictionary Source="Resources/Localization/Dictionary-en-US.xaml"/>
             </ResourceDictionary.MergedDictionaries>
             <FontFamily x:Key="SegoeFluentIcons">pack://application:,,,/;component/Fonts/#Segoe Fluent Icons</FontFamily>
-            
+
+            <!-- set global font family -->
+            <FontFamily x:Key="GlobalFontFamily">Segoe UI Variable, Microsoft YaHei UI, Yu Gothic UI, Malgun Gothic</FontFamily>
+
+            <!-- apply global font family to TextBlocks except ones inside icon controls -->
+            <Style TargetType="TextBlock">
+                <Style.Triggers>
+                    <!-- If not inside a MicaWPF SymbolIcon, apply the GlobalFontFamily -->
+                    <DataTrigger Binding="{Binding RelativeSource={RelativeSource AncestorType={x:Type controls:SymbolIcon}}}" Value="{x:Null}">
+                        <Setter Property="FontFamily" Value="{DynamicResource GlobalFontFamily}" />
+                    </DataTrigger>
+                </Style.Triggers>
+            </Style>
+            <Style TargetType="Control">
+                <Setter Property="FontFamily" Value="{DynamicResource GlobalFontFamily}" />
+            </Style>
+
             <!-- override default background colors -->
             <SolidColorBrush x:Key="NavigationViewContentBackground" Color="Transparent" />
             <SolidColorBrush x:Key="NavigationViewContentGridBorderBrush" Color="Transparent" />
--- a/FluentFlyoutWPF/Classes/LocalizationManager.cs
+++ b/FluentFlyoutWPF/Classes/LocalizationManager.cs
@@ -181,6 +181,17 @@ public static class LocalizationManager
         }
         SettingsManager.Current.FontFamily = fontFamily;
 
+        // apply the font family to the entire application
+        try
+        {
+            var ff = new System.Windows.Media.FontFamily(fontFamily);
+            Application.Current.Resources["GlobalFontFamily"] = ff;
+        }
+        catch (Exception ex)
+        {
+            Logger.Warn(ex, "Failed to apply GlobalFontFamily, keeping SettingsManager value only.");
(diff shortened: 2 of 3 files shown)


unchihugo:
Hi @lindwurm, thank you for looking into this, I wasn't aware that font rendering was spotty!

While I understand that this PR is still a draft, I'd like to add in some context:
1. Based on your screenshots, it looks like the text width helper function for calculating the text width for Next Up and the Taskbar media widget is still using the old font families. 
2. Have you investigated why the viewmodel binding isn't properly working and we have to resort to setting global font family?

Thanks in advance!

lindwurm:
> 1. Based on your screenshots, it looks like the text width helper function for calculating the text width for Next Up and the Taskbar media widget is still using the old font families.

Certainly this font has a narrow character width but I hadn't even considered about that.

> 2. Have you investigated why the viewmodel binding isn't properly working and we have to resort to setting global font family?

To be honest, there isn't enough research. That's also why I've left this PR as a draft (maybe I should have created an issue first).

Since the issue occurs only in certain components,

* It has been confirmed that there is no problem with the fallback order of `_languageFontFamilies[“ja”]`

* but this has not been applied to the affected areas (Media Flyout and the Taskbar Widget)

then I tried to implement that based on hypothesis like:

*  There is no global `FontFamily` resource shared across the entire app, different font resolutions are occurring for each control.

galagyy:
Hello!

I opened a PR fairly recently that resolves this issue to an extent:
> Based on your screenshots, it looks like the text width helper function for calculating the text width for Next Up and the Taskbar media widget is still using the old font families.

The PR uses the following code snippet to get the true size of the text by using a native WPF method. I had an issue with the scrolling not working properly for non-English fonts and had changed it to this to adjust. This means it should still get the size without worrying about which font is being used.
```cs
textBlock.Measure(new Size(double.PositiveInfinity, double.PositiveInfinity));
```

I believe something similar may work in this case; my change in the newer PR only affects scrolling text and doesn't resolve it for non-scrolling text (as is here).

unchihugo:
Hi @lindwurm, I've just written a fix for the string width method in #889, and I've observed that this fixes the issue where the string width was calculated incorrectly with your fix in place. Does it also work on your end if you merge the branch? I'm happy to continue and merge your PR if we see no issues.

lindwurm:
> Hi @lindwurm, I've just written a fix for the string width method in #889, and I've observed that this fixes the issue where the string width was calculated incorrectly with your fix in place. Does it also work on your end if you merge the branch? I'm happy to continue and merge your PR if we see no issues.

It looks good to me, thank you!

Screenshot_20260621_112908_explorer

Screenshot_20260621_113001_explorer

lindwurm:
and after investigating, we found that the issue could be fixed with fewer changes, without using `GlobalFontFamily`. ~~I will update this pull request shortly.~~

*update: I've renamed this branch then GitHub automatically closed this PR.*

## Media Flyout
Since the `DataContext` was set earlier in [MainWindow.xaml.cs](https://github.com/unchihugo/FluentFlyout/blob/c7dcc192fd667100fd1c2a1c155006d88960842a/FluentFlyoutWPF/MainWindow.xaml.cs#L83), there was a possibility that the [RestoreSettings()](https://github.com/unchihugo/FluentFlyout/blob/c7dcc192fd667100fd1c2a1c155006d88960842a/FluentFlyoutWPF/MainWindow.xaml.cs#L83) method in [MainWindow.xaml.cs](https://github.com/unchihugo/FluentFlyout/blob/c7dcc192fd667100fd1c2a1c155006d88960842a/FluentFlyoutWPF/MainWindow.xaml.cs#L133) — which replaces the `SettingsManager.Current` instance itself — would continue to reference the old settings object.

### Solution - Add a rebind to MainWindow.xaml.cs
[Insert](https://github.com/unchihugo/FluentFlyout/blob/c7dcc192fd667100fd1c2a1c155006d88960842a/FluentFlyoutWPF/MainWindow.xaml.cs#L140) `DataContext = SettingsManager.Current;` after `RestoreSettings()` to ensure a connection to the restored instance.

## Taskbar Widget

Just change the `FontFamily` to `{Binding FontFamily}` from a [hard-coded value](https://github.com/unchihugo/FluentFlyout/blob/c7dcc192fd667100fd1c2a1c155006d88960842a/FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml#L12).

PR #858, opened 2026-06-08
Fix: Master and mixer volume sliders not clearing mute state when raised from zero

lechixy:
## Summary

Fixes an issue where the master audio device remained muted after increasing the volume from zero.

## Motivation

When the master volume was set to 0, `IsMasterMuted` was correctly set to `true`. However, increasing the volume afterwards did not reset `IsMasterMuted` back to `false`, causing the mute icon to remain visible and audio output to stay muted.

## Type of Change

- [x] Bug fix

## What Changed

- Added missing logic to set `IsMasterMuted` to `false` when the master volume is greater than zero.
- Ensured the mute state always stays synchronized with the current master volume level.

## Additional Information

Here is how it was looking before:

image

Before this change:
- Setting volume to 0 muted the device.
- Increasing the volume afterwards did not unmute it.

After this change:
- The mute state is automatically updated whenever the volume changes.
- Audio output and mute icon behavior now correctly reflect the current volume level.

## Checklist

- [ ] AI tools were used (if yes, I reviewed and fully understand the changes myself).

### DIFF
--- a/FluentFlyoutWPF/Models/AudioSessionModel.cs
+++ b/FluentFlyoutWPF/Models/AudioSessionModel.cs
@@ -51,6 +51,9 @@ public partial class AudioSessionModel : ObservableObject
         if (Volume == 0f)
         {
             IsMuted = true;
+        } else
+        {
+            IsMuted = false;
         }
     }
 
--- a/FluentFlyoutWPF/Models/AudioSessionModel.cs
+++ b/FluentFlyoutWPF/Models/AudioSessionModel.cs
@@ -51,7 +51,8 @@ public partial class AudioSessionModel : ObservableObject
         if (Volume == 0f)
         {
             IsMuted = true;
-        } else
+        }
+        else
         {
             IsMuted = false;
         }
--- a/FluentFlyoutWPF/ViewModels/VolumeMixerViewModel.cs
+++ b/FluentFlyoutWPF/ViewModels/VolumeMixerViewModel.cs
@@ -102,6 +102,9 @@ public partial class VolumeMixerViewModel : ObservableObject, IDisposable
         if (MasterVolume == 0f)
         {
             IsMasterMuted = true;
+        } else
+        {
+            IsMasterMuted = false;
         }
     }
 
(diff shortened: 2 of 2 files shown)


unchihugo:
Hi @lechixy, thank you for contributing, I appreciate that you've taken your time looking into this! 

While the functionality is good, the current behavior is actually intended: it aims to mimic how Windows 11 itself handles volume sliders. If you open up the native volume mixer for example (Win + CTRL + V) the sliders react in a similar way. Unfortunately, I'll have to close the PR for now due to that reason.

To prevent things like this happening in the future, it can be smart to open up an issue for this first before creating a PR. Again, I do appreciate your effort and I'm always willing to reopen if there's demand!

lechixy:
> Hi @lechixy, thank you for contributing, I appreciate that you've taken your time looking into this!
> 
> While the functionality is good, the current behavior is actually intended: it aims to mimic how Windows 11 itself handles volume sliders. If you open up the native volume mixer for example (Win + CTRL + V) the sliders react in a similar way. Unfortunately, I'll have to close the PR for now due to that reason.
> 
> To prevent things like this happening in the future, it can be smart to open up an issue for this first before creating a PR. Again, I do appreciate your effort and I'm always willing to reopen if there's demand!

Thanks for the clarification.

I may have explained the issue poorly in the PR description.

The reason I submitted this change was not to make volume level and mute state behave the same thing, but because `IsMasterMuted` remained `true` after increasing the volume from zero again.

In my testing:

- Set volume to 50%
- Drag slider to 0%
- Drag slider back to 30%

Windows immediately restores audio output at 30%.

However, in the application, audio remained muted and the mute icon was still shown because `IsMasterMuted` stayed `true` after the volume was increased again.

The added `else` only clears the stale mute state when the volume is no longer zero, matching the behavior I observed in Windows.

If I'm misunderstanding the intended behavior or if there's another part of the code handling this, please let me know. Thanks for your time.

unchihugo:
Actually, you're right that Windows unmutes after dragging again. I just tested myself and it seems like I was wrong!

unchihugo (dismissed):

unchihugo:
Just run a quick `dotnet format` command on the edited files and we should be ready to merge!

lechixy:
> Actually, you're right that Windows unmutes after dragging again. I just tested myself and it seems like I was wrong!

No worries mate ❤️😊

unchihugo (approved):

Issue #278, opened 2026-01-11
[FEATURE REQUEST] Progressbar, Media title scrolling, Enlarged album cover


zeroma25:
**Purpose:**  
A few small enhancements.

**Description:**  
1. Add a progressbar to the widget. I think in this case it is better to add a custom fixed widget width.
2. Scrolling a long media title that does not fit in the widget during playback.
3. Enlarged album cover on hover.
4. The option "Click to open Media Flyout when media is Present" should work even when Media Flyout is turned off.

**Additional context:**  
![Image](https://github.com/user-attachments/assets/8fb783b1-71fd-4c88-9b2d-49fdff238272)

unchihugo:
Hi, thank you for the suggestions!

I've already started to work on the album art hover tooltip, you can expect that in the next update (if everything goes right). Just wondering, did you create the UI in the demo video yourself?

zeroma25:
@unchihugo This is a Windhawk mod that I edited for myself using Gemini AI.

PR #874, opened 2026-06-15
feat: add scrolling text to taskbar widget (#576, #278)

### LINKED ISSUE #576: [FEATURE REQUEST] Scrolling text on taskbar widget
### Description

A scrolling text option would be nice for when titles are too long and get cut off.
Something like this:

https://github.com/user-attachments/assets/5fde657a-d5c1-488b-b3d4-6a8657bd9d52

### Additional context

_No response_

galagyy:
## Summary
This PR introduces an app scrolling feature. I saw multiple feature requests with this idea (#278 and #576) and liked the idea myself.

This feature allows the user enable scrolling text for their taskbar widget. Users may adjust the scrolling (default or continuous), speed (how fast to move text, in pixels), and whether they want the artist's name to also scroll. 

## Motivation
Aims to close #576 and parts of #278.

## Type of Change
- [x] Feature
- [ ] Bug fix
- [x] Style (formatting, naming)
- [ ] Other

## What Changed
The taskbar widget has changed to accompany the following features:
- Added option to enable scrolling for text that does not fit the widget
- Added two scrolling modes: default scrolling (akin to Spotify) or continuous scrolling
- Added option to configure the speed at which the scrolling should occur (in pixels per second)
- Added option to toggle whether the artist or song name should be scrolling

## Additional Information
Here are screenshots for each "mode:"

### UI Settings
Settings Screenshot

### Default Mode
The default mode as shown on Spotify, with text "ping ponging." (I think that's what they call it)
Default Scrolling

### Infinite Scrolling
Infinite Scrolling gif

NOTE: The GIFs are not perfectly cut; the scrolling in the actual code is smoother and loops properly. The GIFs are also in 24 FPS.

## Checklist
- [x] AI tools were used (Used to verify code is bug-free and to standard).

This is my second pull request with C# so hopefully the code should be cleaner than last time! Thank you for your help in the previous PR.


galagyy:
**Notes:**
* I have also added a small transparency gradient to the left and right edges of the text to make the scrolling more smooth. This only displays when the "scrolling text" option is enabled.
* I have verified the scrolling works with artists and/or songs in another language, such as Japanese and Southeast Asian languages. I cannot say it works with ALL languages though, as that might need further testing.

galagyy:
I wasn't sure if there was a consistent naming scheme for PRs so please excuse the name changes!

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -202,19 +226,22 @@ private void Grid_MouseLeftButtonUp(object sender, MouseButtonEventArgs e)
 
     public (double logicalWidth, double logicalHeight) CalculateSize(double dpiScale)
     {
-        // calculate widget width - use cached values if text hasn't changed
```
unchihugo, line 205: Just wondering here, why is the implementation different and comment removed? A lot of additional lines get added since this check has been changed, but I can't see the benefit immediately.

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -227,32 +254,181 @@ private void Grid_MouseLeftButtonUp(object sender, MouseButtonEventArgs e)
         }
         else
         {
-            logicalWidth = Math.Max(_cachedTitleWidth, _cachedArtistWidth) + 55; // add margin for cover image
```
unchihugo, line 230: I think we should keep this comment in as it clarified what +55 means

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -227,32 +254,181 @@ private void Grid_MouseLeftButtonUp(object sender, MouseButtonEventArgs e)
...
         }
 
-        SongTitle.Width = Math.Max(logicalWidth - 58, 0);
-        SongArtist.Width = Math.Max(logicalWidth - 58, 0);
+        double newTitleContainerWidth = Math.Max(logicalWidth - 58, 0);
+        double newArtistContainerWidth = Math.Max(logicalWidth - 58, 0);
+
+        double availableTextWidth = Math.Max(maxLogicalWidth - 58, 0);
+
+        bool scrollingTitleSetting = SettingsManager.Current.TaskbarWidgetScrollingTitleText;
```
unchihugo: While this line and the code beyond behaves similar to what we've done in `MainWindow.cs` (Media Flyout), with the taskbar widget we've opted for a more dynamic refreshing method with UpdateTaskbar(), called in `UserSettings.cs` whenever a setting changes (take a look at methods that call UpdateTaskbar() in there). This is much more readable and makes caching mostly unnecessary in here, could you take a look at that implementation?

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -227,32 +254,181 @@ private void Grid_MouseLeftButtonUp(object sender, MouseButtonEventArgs e)
...
             logicalWidth += (int)(102);
         }
 
-
         double logicalHeight = 40; // default height
 
         return (logicalWidth, logicalHeight);
     }
 
+    private void UpdateMarquee(System.Windows.Controls.TextBlock textBlock, Canvas container, double textWidth, double availableWidth, bool isEnabled, int speed, bool loopForever)
```
unchihugo: This method looks great! In the future, I think we should definitely add this method into a helper/util class so we can apply this effect to all other textboxes that could benefit from this as well :)

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -227,32 +254,181 @@ private void Grid_MouseLeftButtonUp(object sender, MouseButtonEventArgs e)
...
+        double newArtistContainerWidth = Math.Max(logicalWidth - 58, 0);
+
+        double availableTextWidth = Math.Max(maxLogicalWidth - 58, 0);
+
+        bool scrollingTitleSetting = SettingsManager.Current.TaskbarWidgetScrollingTitleText;
+        bool scrollingArtistSetting = SettingsManager.Current.TaskbarWidgetScrollingArtistText;
+        int scrollingSpeed = SettingsManager.Current.TaskbarWidgetScrollingTextSpeed;
+        bool scrollingLoop = SettingsManager.Current.TaskbarWidgetScrollingTextLoopForever;
+
+        bool settingsChanged = _lastScrollingTitleSetting != scrollingTitleSetting ||
```
unchihugo: Most of the following code works quite similarly to each other, if you implement this into UpdateTaskbar(), you could definitely consolidate it to simply just update both textboxes if the setting is turned on!

unchihugo (changes requested):
Hi @galagyy, I've only looked through the code and your videos so far but it does look really great! I've written up some things that I think should be changed up before we merge this one as well, since it's a larger PR. Feel free to comment on changes that you believe are necessary!

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
galagyy: That's my bad; I'll add back all the comments as I accidentally removed them while testing something.

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
galagyy: I've gone ahead and removed the old implementation in favor of this new, improved version.

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
galagyy: Thank you! I've made an attempt to also "decouple" it so that it can easily become a standalone method in the future.

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
galagyy: I've connected them both via the `UpdateMarquees()` method so that they may both be updated together as you mentioned!

unchihugo:
Looks better! Just wondering though why you've changed the implementation of currentArtist and currentTitle? Is the new caching method necessary for the new code?

FluentFlyoutWPF/ViewModels/UserSettings.cs
```diff
@@ -885,6 +938,38 @@ private void UpdateTaskbar()
...
+        RefreshMarquees();
+    }
+
+    partial void OnTaskbarWidgetScrollingTextSpeedChanged(int oldValue, int newValue)
+    {
+        if (oldValue == newValue || _initializing) return;
+        RefreshMarquees();
+    }
+
+    private void RefreshMarquees()
```
unchihugo: Could you rename this to `UpdateTaskbarMarquees` just in case we start adding marquees to other elements as well?

FluentFlyoutWPF/Controls/TaskbarWidgetControl.xaml.cs
```diff
@@ -231,28 +246,146 @@ private void Grid_MouseLeftButtonUp(object sender, MouseButtonEventArgs e)
...
 
         return (logicalWidth, logicalHeight);
     }
 
+    public void UpdateMarquees()
+    {
+        double titleAvailableWidth = double.IsNaN(SongTitleContainer.Width) ? 0 : SongTitleContainer.Width;
+        double artistAvailableWidth = double.IsNaN(SongArtistContainer.Width) ? 0 : SongArtistContainer.Width;
+
+        UpdateMarquee(SongTitle, SongTitleContainer, _cachedTitleWidth, titleAvailableWidth, SettingsManager.Current.TaskbarWidgetScrollingTitleText);
```
unchihugo: Let's consolidate the setting "TaskbarWidgetScrollingTitleText" and "TaskbarWidgetScrollingArtistText" into one "TaskbarWidgetScrollingEnabled" option, since I'd think most users would want both on/off at the same time :)

unchihugo (changes requested):
The changes make the code much cleaner already. Just a few more steps!

galagyy:
Hello!

Thanks for the prompt response, with regards to the changes to `currentArtist` and `currentTitle`, I've changed it so that we're querying our own variables instead of directly calling the UI elements. Querying the UI during calculations led to CPU spikes on occasion which was resolved by doing this.

I also added flags to guard `UpdateMarquees()` as without it would be called repeatedly causing issues.

I'll address the other issues shortly!

galagyy:
I've moved the `55`/`58` to a constant as it was serving similar to a "magic number." Changing that number without changing the number in other places would break the scrolling system.

unchihugo:
> I've moved the `55`/`58` to a constant as it was serving similar to a "magic number." Changing that number without changing the number in other places would break the scrolling system.

Could you adjust _coverImageMargin to be 55, and in line 251 and 252 do _coverImageMargin + 3? Not sure why it was +3 before but it must've had a reason (which proves your point of having magic numbers, I'll definitely think of that in the future)

galagyy:
> Could you adjust _coverImageMargin to be 55, and in line 251 and 252 do _coverImageMargin + 3? Not sure why it was +3 before but it must've had a reason (which proves your point of having magic numbers, I'll definitely think of that in the future)

Will do, also with the latest change I observed the pause icon setting broke so I'm looking into that.

galagyy:
I resolved the pause icon issue, and I resorted to keeping them both as `55` as they are meant to be the same value.

unchihugo:
> I resolved the pause icon issue, and I resorted to keeping them both as `55` as they are meant to be the same value.

are you able to change line 251 and 252 to the variable + 3 as well like the previous version?

galagyy:
I tried doing that but it doesn't work properly; we get a negative number from that on occasion and this condition is always triggered:
```
textWidth > availableWidth
```

This means that text scrolls even if the size of the text is small enough to fit in the area, which is not good behavior.

I can try changing the condition, but it appears to work fine with both as 55, as the +3 margin wasn't too important.

unchihugo:
Alright, that makes sense. Have you checked if the text box still trims properly if the marquee is disabled since the margins are changed (when the text is just small enough to fit the area & when it's just a bit too big)?

galagyy:
I've tested the ellipses returning for text smaller than the container, about the exact size of the container, and larger than the container. 

I believe it works properly; if it should exactly fit the entire container it does appear to do so.  I've also tested with songs which have the artist and/or song name in JP and they appear to also work properly.

I wasn't able to find a song that is an exact fit though (fitting the entire size of the container exactly). The code block responsible for this uses one of WPF's methods so I believe it should be reliable:
```cs
else
{
    transform.BeginAnimation(TranslateTransform.XProperty, null);
    transform.X = 0;
    textBlock.Text = isTitle ? _actualTitle : _actualArtist;
    textBlock.Width = containerWidth;
    textBlock.TextTrimming = TextTrimming.CharacterEllipsis; // Should be reliable in this case
    container.OpacityMask = null;
}
```

unchihugo:
Hi @galagyy, I've tested the latest version and it looks incredibly good! During testing I noticed that there's a transparency gradient even when the text is all the way to the left (also observable in your gif, where "G" is semi-transparent):
> image
Do you think there's a way to have the transparency effect disabled when that's the case?

Also, just to clarify, does Spotify provide options for continuous scrolling and the scrolling speed? Great work so far, I really appreciate it :)

galagyy:
Hello!

Thank you for the prompt response & I appreciate your compliment!

With regards to the text, I'll get that disabled for the "ping pong" like mechanic. I had that there originally mainly for the continuous scrolling but I admittedly forgot to implement it properly for the default mechanic.

Regarding Spotify, I don't believe they provide options for the user to customize the scrolling speed of the text, I added that to give the user more options. For the continuous scrolling I implemented that since I saw that being followed in other media apps (as in the video in #576) and thought it looked nice.

The Spotify side bar will have the song name follow the default methodology if the song name exceeds the limit:

Spotify Example

galagyy:
I should have that issue fixed now, I added a few keyframes and adjusted the code a bit so that the gradient is not present at the start or end of an animation on the left-hand side. If the setting is on "continuous scrolling" then it maintains the current behavior. I also cleared the color animations when unused as I would see small spikes from them even if they weren't being used.

It's also did some tests and found it's worth noting that text in pure Japanese tends to appear smaller but actually triggers the scrolling mechanic earlier due to them being full-width/square characters.

unchihugo:
> It's also did some tests and found it's worth noting that text in pure Japanese tends to appear smaller but actually triggers the scrolling mechanic earlier due to them being full-width/square characters.

I've also noticed something similar and I believe this is because the text measuring method is a little outdated. I'm working on a patch to fix this. I'll check your changes when I'm available!

galagyy:
Thank you for the prompt response and help, and no worries! I will be away for a day or two before being able to work on this again.

galagyy:
Also, regarding the Japanese text, I was able to resolve that by adding a small margin in the if statement that decides whether to scroll or not. I didn't push it yet to respect the patch that you were working on.
```cs
// Line 310 of TaskbarWidgetControl.xaml.cs
if (isEnabled && textWidth > availableWidth + 0.5 && containerWidth > 0 && !double.IsNaN(containerWidth))
```

I found that there are slight measurement issues when measuring Japanese and possible other languages; maybe this information could aid with the new measuring method. For example, when a Japanese song was entered, the following were the values of certain variables:
```
Calculated TextWidth: 112.00000000000001
AvailableWidth: 112
```

I believe this stems from computers using base-2, and may be causing issues elsewhere potentially.

unchihugo:
@galagyy just wondering, is the app language English or another language for you when you observe issues with the Japanese text? Do you still see errors in width calculation after merging master with the new string width changes?

galagyy:
Hello,

I've been using English as the app language when the issue was observed. With regards to the width calculation, the issue appears to still persist (I think it's still a math error like my previous comment). The small margin idea earlier does still appear to fix it though, but it's a bit unclean.

unchihugo:
Alright, that narrows the issue. What happens if you remove or reduce +8 in the following line? It might be that? https://github.com/unchihugo/FluentFlyout/blob/dee4e3f29ebb3b4ab2ab456a8a4c42442251e4d6/FluentFlyoutWPF/Classes/Utils/StringWidth.cs#L37

galagyy:
Removing the `+ 8` still has that issue present. I think the very abnormal value (the .000001 earlier) stems from how the computer itself handles numbers (not being able to store certain fractions and stuff properly).

unchihugo:
Oh, if the floating point error is causing the issue, can you try rounding the TextWidth value after measuring?

galagyy:
The rounding fixed it, it no longer starts scrolling when not needed!

I kept the `+ 8` that was outlined earlier as it did not affect anything here.

unchihugo (changes requested):
Hi @galagyy, sorry for the late response. I've tested the new functionality and it seems like the text sometimes scrolls even if it fits inside the maximum widget width, e.g. if the title is 50px wide, on some occasions it still might scroll. 

This has to do with padding/margins with the + 8 in the measure method: https://github.com/unchihugo/FluentFlyout/blob/dee4e3f29ebb3b4ab2ab456a8a4c42442251e4d6/FluentFlyoutWPF/Classes/Utils/StringWidth.cs#L37

If you remove the + 8 from this line, I believe this stops happening, though we would have to add an 8px margin in the XAML code to make up for this. Can you reproduce this on your side?

I have also noticed that there's quite a lot of magic numbers in the UpdateMarquee() method, do you think you could comment the numbers for better documentation?

Many thanks in advance for the help :)

FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
```diff
@@ -263,6 +263,14 @@ You can help contribute translations on</System:String>
     <System:String x:Key="AdvancedSettingsTitle">Advanced Settings</System:String>
     <System:String x:Key="LegacyTaskbarWidthTitle">Legacy Taskbar Width System</System:String>
     <System:String x:Key="LegacyTaskbarWidthDescription">Use the legacy Taskbar width system for Taskbar Widgets</System:String>
+    <System:String x:Key="TaskbarWidgetScrollingTextTitle">Scroll Text</System:String>
```
unchihugo: Could you reword this to "Scrolling Text"?

FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
```diff
@@ -263,6 +263,14 @@ You can help contribute translations on</System:String>
...
+    <System:String x:Key="TaskbarWidgetScrollingTitle">Enable Scrolling</System:String>
+    <System:String x:Key="TaskbarWidgetScrollingTitleDescription">Enable scrolling for the artist and song name in the taskbar widget.</System:String>
+    <System:String x:Key="TaskbarWidgetScrollingTextDescription">Automatically scroll long text that doesn't fit in the widget</System:String>
```
unchihugo: Could you rephrase this to "Automatically scroll long text that doesn't fit"? This is so that we can use this sentence in the future for other flyouts as well without having to add new strings to the dictionary.

FluentFlyoutWPF/Resources/Localization/Dictionary-en-US.xaml
```diff
@@ -263,6 +263,14 @@ You can help contribute translations on</System:String>
     <System:String x:Key="AdvancedSettingsTitle">Advanced Settings</System:String>
     <System:String x:Key="LegacyTaskbarWidthTitle">Legacy Taskbar Width System</System:String>
     <System:String x:Key="LegacyTaskbarWidthDescription">Use the legacy Taskbar width system for Taskbar Widgets</System:String>
+    <System:String x:Key="TaskbarWidgetScrollingTextTitle">Scroll Text</System:String>
+    <System:String x:Key="TaskbarWidgetScrollingTitle">Enable Scrolling</System:String>
+    <System:String x:Key="TaskbarWidgetScrollingTitleDescription">Enable scrolling for the artist and song name in the taskbar widget.</System:String>
```
unchihugo: I think we should remove this string and use the "TaskbarWidgetScrollingTextDescription" string in this place in the XAML instead, and remove the description from the "Enable Scrolling" toggle. Just thinking of translation redundancy

FluentFlyoutWPF/Pages/TaskbarWidgetPage.xaml
```diff
@@ -259,7 +260,49 @@
                 </ui:CardControl.Header>
                 <controls:ToggleSwitch IsChecked="{Binding TaskbarWidgetAnimated, Mode=TwoWay}" IsEnabled="{Binding IsPremiumUnlocked}" />
             </ui:CardControl>
-            <ui:CardControl Grid.Row="14" Icon="{ui:SymbolIcon Pause24}" Margin="0,0,0,76">
+
+            <ui:CardExpander Grid.Row="14" Icon="{ui:SymbolIcon TextSortAscending24}" ContentPadding="6" Margin="0,0,0,3">
```
unchihugo: I think using the "Switch24" icon would make a bit more sense here, what do you think?

unchihugo (commented):
Also, leaving some comments regarding the dictionary and page. Incredibly sorry for the inconvenience, but as mentioned before, pushing dictionary changes to master means our translator contributors have to work with them, so it’s best to get it right the first time. :)

galagyy:
No worries for the late response! I should have them fixed by today/tomorrow.

FluentFlyoutWPF/Pages/TaskbarWidgetPage.xaml
galagyy: Yeah that makes sense! I'll change it to that.

galagyy: I was thinking of something else; I wasn't able to find a "Switch24" icon, so I changed it to the icon below for now:

Scrolling Text Icon Image

unchihugo:
Thank you for the quick changes! I'll review it later today

unchihugo (approved):
Looks good now, thank you!

galagyy:
No problem! 

Do you have any other issues that you would like resolved or feature requests that you would like added? I saw a few issues and feature requests a bit earlier that I would like to add, but was wondering if you had any preference.

unchihugo:
Not really any smaller issues, but I'd like to adapt the marquee method to work with any text block! We'll have to move the method to its own class, and I'm definitely looking to track the text states for changes in that class instead of having to introduce text state tracking in every text block that uses the marquee in the future.

I'm wondering though, what do you feel like working on next?

galagyy:
The marquee point makes sense, I was thinking an implementation similar to this:
- A class that wraps a `(Textblock, Canvas)` pair with it's own cached width, opacity mask, and last-animation parameters unique to itself
- Expose something like `Attach(TextBlock, Canvas)` and `Update(string text, double availibleWidth, bool enabled, int speed, bool loopForever)` so that any callee owns an instance per text block
- `TaskbarWidgetControl` would have two instances (title and artist) instead of parallel fields, any any future text block anywhere else in the app can use the same behavior for free.

I haven't fully thought the implementation through, so there may be a few holes; I was last thinking about how this might be a bit resource intensive and if there was a better approach.

With regards to what to work on next, I'm fine with taking anything really as I don't really have a strong preference for or against anything. I like working on issues that have the "enhancement" tag more as they involve adding new features, but I'm also fine with working on bugs. I had noticed a small bug with `ControlSession` earlier today that I aim to open a PR for soon.

unchihugo:
An object class system makes a lot of sense actually. I was thinking a static utility class, but I like both ideas equally.

> I was last thinking about how this might be a bit resource intensive and if there was a better approach.

I'm sure it would be fine!

You can work on whatever you want, I appreciate your help :)

galagyy:
Thanks for the prompt reply! I'll work on getting the Marquee system working better soon with the implementation that I had outlined earlier.

> You can work on whatever you want, I appreciate your help :)

I saw all the open issues/feature requests and don't really have a preference for the ones there, so I was wondering if you had any preference in what would be worthwhile doing first.

Thanks again!

unchihugo:
Actually, I did have in mind adding a copy/paste flyout that displays when the user copied or pasted! 

Shouldn't be too much effort but I'm still stuck on how it should look UI wise. If we make it look exactly like the lock keys flyout, we'll most likely get away with reusing the LockWindow for this.

The easiest way to do that would probably be:
1. add a copy/paste hook in the existing MainWindow hooks code, and possibly NativeMethods (we should definitely change the messy hook system in the future)
2. Call LockWindow, specify that it's a copy or paste event
3. adjust LockWindow UI accordingly
4. Add a new Clipboard flyout page in the settings with toggles

Though again, that is if we want the clipboard flyout to look the same as lock keys. I've also thought about it just being a little status icon, or indicator, or glow on the taskbar, but I'm stuck on it. Do you think you could add to this idea?

galagyy:
Thanks for the prompt response! The copy/paste flyout does seem like a good idea; reusing `LockWindow` does make sense given it's the same category of event as the latter. 

I think idea-wise maybe checking if the clipboard content actually changed is a better metric for whether the UI is trigged, though this may be confusing for some users. I've noticed that several users on Windows, including myself, sometimes press `Ctrl + C` but don't actually see the content copied to the clipboard; only after pressing it a few times does the content only change.

I think another cool feature could be looking at how the `WM_CLIPBOARDUPDATE` event fires on file copies as well, we could check the clipboard via `Clipboard.ContainsFileDropList()` and show something like "1 file copied" or "Text Copied," depending on the type of content being changed.

I also like the idea of a status icon/indicator/glow for the flyout; I'll spend some time looking at how those work on Windows and experimenting a bit to see what would be best.

PR #906, opened 2026-06-29
Fix: Clear Uuid and IsStoreVersion on export

### LINKED ISSUE #905: [FEATURE REQUEST] Clear `Uuid` and `IsStoreVersion` when exporting settings through the app
### Description

Clear the `Uuid` and `IsStoreVersion` fields when exporting settings through the app, since they could be conflicting and include info that users might not want to share. When importing, these two fields should be ignored, and the user's current values should overwrite them.

### Additional context

Import settings method:
https://github.com/unchihugo/FluentFlyout/blob/05896ed46d54c3014009ff8f19c10cd55b4800ca/FluentFlyoutWPF/Classes/Settings/SettingsManager.cs#L57

Export settings method:
https://github.com/unchihugo/FluentFlyout/blob/05896ed46d54c3014009ff8f19c10cd55b4800ca/FluentFlyoutWPF/Classes/Settings/SettingsManager.cs#L109

Both methods are used for both normal saving/loading and export saving/loading. The only difference is that exporting/importing uses the `filePath` parameter, whereas normal saving/loading doesn't.

Yasinnet0:
Fixes #905

This PR clears the `Uuid` and `IsStoreVersion` fields when exporting settings so users don't share private IDs. When importing, these two fields are ignored from the file, and the user's current values are retained.

Changes:
- Uses `XmlAttributeOverrides` during export to ignore these fields without triggering the observable property background save.
- Caches the custom serializer to prevent memory leaks.
- Retains the active `Uuid` and `IsStoreVersion` during import by overwriting the deserialized settings.

### DIFF
--- a/FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
+++ b/FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
@@ -22,6 +22,21 @@ public class SettingsManager
     );
 
     private static UserSettings? _current;
+    private static XmlSerializer? _exportSerializer;
+
+    private static XmlSerializer GetExportSerializer()
+    {
+        if (_exportSerializer == null)
+        {
+            XmlAttributeOverrides overrides = new XmlAttributeOverrides();
+            XmlAttributes ignoreAttrs = new XmlAttributes();
+            ignoreAttrs.XmlIgnore = true;
+            overrides.Add(typeof(UserSettings), "Uuid", ignoreAttrs);
+            overrides.Add(typeof(UserSettings), "IsStoreVersion", ignoreAttrs);
+            _exportSerializer = new XmlSerializer(typeof(UserSettings), overrides);
+        }
+        return _exportSerializer;
+    }
 
     private static bool DeserializeSettings(string filePath, out UserSettings? settings)
     {
@@ -110,8 +125,6 @@ public class SettingsManager
         return _current;
     }
 
-    private static XmlSerializer? _exportSerializer;
-
     /// <summary>
     /// Saves the app settings to the settings file.
     /// </summary>
@@ -139,16 +152,7 @@ public class SettingsManager
                     XmlSerializer xmlSerializer;
                     if (isExport)
                     {
-                        if (_exportSerializer == null)
-                        {
-                            XmlAttributeOverrides overrides = new XmlAttributeOverrides();
-                            XmlAttributes ignoreAttrs = new XmlAttributes();
-                            ignoreAttrs.XmlIgnore = true;
-                            overrides.Add(typeof(UserSettings), "Uuid", ignoreAttrs);
-                            overrides.Add(typeof(UserSettings), "IsStoreVersion", ignoreAttrs);
-                            _exportSerializer = new XmlSerializer(typeof(UserSettings), overrides);
-                        }
-                        xmlSerializer = _exportSerializer;
+                        xmlSerializer = GetExportSerializer();
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
```diff
@@ -56,13 +56,20 @@ public static UserSettings Current
...
         filePath ??= SettingsFilePath;
         string backupPath = filePath + ".bak";
 
         try
         {
             if (DeserializeSettings(filePath, out var loadedSettings) && loadedSettings != null)
             {
+                if (isImport && _current != null)
+                {
+                    loadedSettings.Uuid = _current.Uuid;
```
unchihugo, line 84: What happens here if `loadedSettings.Uuid` doesn't exist?

FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
```diff
@@ -103,11 +110,14 @@ public static UserSettings RestoreSettings(string? filePath = null)
         return _current;
     }
 
+    private static XmlSerializer? _exportSerializer;
```
unchihugo: Could you move this variable to the start of this class? It's basic C# convention :)

FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
```diff
@@ -126,7 +136,24 @@ public static void SaveSettings(string? filePath = null)
 
                 using (var writer = new StreamWriter(tempPath, false))
                 {
-                    XmlSerializer xmlSerializer = new(typeof(UserSettings));
+                    XmlSerializer xmlSerializer;
+                    if (isExport)
```
unchihugo, line 153: Could you wrap the if block of isExport in its own private static method?

unchihugo (changes requested):
Hi @Yasinnet0, thank you for the contribution!

I've left a few comments on your code, could you resolve them before we merge?

FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
Yasinnet0: I added those lines as a safeguard for when users import files. If a user imports a settings file from an
older version of the app (or from another user's PC), that file might still contain an old  Uuid  and IsStoreVersion . These lines act as a shield to guarantee the app's current  Uuid  and Store status don't get accidentally overwritten by whatever is in the imported file.

As for what happens if it doesn't exist in the file: Since  Uuid  is a  Guid it can't technically
be null. If the  <Uuid>  tag is completely missing from the XML, the deserializer just gives it a default value,
which we then safely overwrite with the active  _current.Uuid  on the very next line. So there is zero risk of an error or crash

FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
Yasinnet0: oh my bad sorry for that fixed it in the new commit😅

FluentFlyoutWPF/Classes/Settings/SettingsManager.cs
unchihugo: I see, both are completely safe if the property itself doesn't exist?

Yasinnet0: > I see, both are completely safe if the property itself doesn't exist?

  If they don't exist in the imported file at all, the  XmlSerializer  just gives them their standard default values
  (a random  Guid  and  false for the storeversion ) It reads the file perfectly without throwing any errors.

  Then, our code instantly catches those defaults and safely overwrites them with the app's real  _current  values. So
  there is zero risk of an error, and it completely protects the user's true UUID and Store License from being
  overwritten by imported files

unchihugo (approved):
Thank you for the contribution!

PR #914, opened 2026-06-29
feat: add global app search

Yasinnet0:
Added a new global app search feature to make it easier for users to find specific settings and sections quickly.

- Placed a unified search bar in the TitleBar so it's accessible from everywhere in the app.
- Search queries index all items from the navigation view (like System, Media, Taskbar, etc.) and deep-link directly to them.
- Fixed layout properties to ensure the new search bar doesn't overlap with the main app content.


unchihugo (changes requested):
Looks great, nice work! I love the heartbeat highlighting idea, and I'm definitely looking forward to having this merged since we've been expanding our settings.

Just wondering, how performant is the `AddDeepLink()` method (since it tries finding resources for every call)? 

Also, I definitely feel like there's an easier way to make all options searchable instead of adding a unique name to each setting, then manually adding them starting at line 174: https://github.com/unchihugo/FluentFlyout/pull/914/changes#diff-fa57c105be0e96f1831f0a78a9a8007180b8992354f43085e3afef50f05c36cdR174

Could you investigate this before we merge? thank you!

Yasinnet0:
> Looks great, nice work! I love the heartbeat highlighting idea, and I'm definitely looking forward to having this merged since we've been expanding our settings.
> 
> Just wondering, how performant is the `AddDeepLink()` method (since it tries finding resources for every call)?
> 
> Also, I definitely feel like there's an easier way to make all options searchable instead of adding a unique name to each setting, then manually adding them starting at line 174: https://github.com/unchihugo/FluentFlyout/pull/914/changes#diff-fa57c105be0e96f1831f0a78a9a8007180b8992354f43085e3afef50f05c36cdR174
> 
> Could you investigate this before we merge? thank you!

yes absolutely i'm working on it right now i'm going to do a commit to fix the merge conflict and then will look at the other stuff

Yasinnet0:
> Just wondering, how performant is the `AddDeepLink()` method (since it tries finding resources for every call)?

The  AddDeepLink()  method is actually very performant and has almost zero overhead

Here is exactly how it works and why we use  TryFindResource :

1. why we use TryFindResource: we use it to instantly grab the correct language translation. by passing it a key
(like  "TaskbarWidgetEnabledTitle" ), it fetches the actual human readable text for the users current language
(like  "Taskbar Position"  in English or  "Position de la barre des tâches"  in French). This means our search box
is fully localized, and users can search using their native language
2. Memory speed:  Application.Current.TryFindResource()  doesn't scan the hard drive. WPF ResourceDictionaries are
backed by highly optimized Hashtables in memory. Looking up ~60 text keys takes less than 1 millisecond on modern
cpus and doesn't block the UI thread at all.
3. Typing performance: It only runs exactly once in the background when the  SettingsWindow  initially loads.
because it builds and caches this localized list beforehand there are no resource lookups happening while the user
is actually typing. Typing latency is basically zero

> Also, I definitely feel like there's an easier way to make all options searchable instead of adding a unique name to each setting, then manually adding them starting at line 174: https://github.com/unchihugo/FluentFlyout/pull/914/changes#diff-fa57c105be0e96f1831f0a78a9a8007180b8992354f43085e3afef50f05c36cdR174

Regarding an easier way to make options searchable.
You are right that manually adding them to the list is tedious. Because WPF-UI lazily loads its NavigationView pages
those pages literally do not exist in the visual tree or in memory until the user clicks their tab. If we wanted to dynamically build the search list by auto-scanning the UI we would be forced to instantiate and load every single page the second the app boots up. This would defeat the performance benefits of lazy loading and cause a spike in memory and startup time. Maintaining a hardcoded registry via AddDeepLink is necessary to keep the app fast while giving the search box a list of suggestions.

We can clean up the XAML though. Adding x:Name to every single Card in the XAML files is messy. If you want I can
write a FindChildByText helper. Instead of relying on x:Name to find the card the code will just scan the page after
it loads for the exact localized text we are already translating and then highlight its parent Card. but i suggest keeping it because if we remove it we have to scan for it through the page but with x:Name it just knows where to look


sorry for lot of explanation 😅

unchihugo:
Thanks for the explanation. I still believe there's a better solution to hard coding every setting though, have you looked into adding Tag="Indexable" properties and indexing them by traversing the logical tree instead of visual tree?

Yasinnet0:
> Thanks for the explanation. I still believe there's a better solution to hard coding every setting though, have you looked into adding Tag="Indexable" properties and indexing them by traversing the logical tree instead of visual tree?

I considered traversing the Logical Tree, but there are two major issues with that approach in WPF:
  1. To traverse the Logical Tree of a page, that page still has to be instantiated in memory first ( new
  TaskbarWidgetPage() ). If we instantiate every page at startup just to index them, we completely destroy the memory
  benefits of lazy loading.
  2. If we tag elements with  Tag="Indexable" , we still have to figure out what text to show in the search box.
  Because our strings are  DynamicResource  bindings, trying to programmatically extract the ResourceKey out of a
  logical tree element's property and i personally think its not good to do it 
  
but if you still think its a good idea to do it i can go ahead and do it

unchihugo:
Hi @Yasinnet0, I've been looking into this issue and I believe we can make this work in a different way so we don't have to hardcode every searchable option in code. What I think would work much better:

1. Remove the hardcoded `AddDeepLink()` calls
2. Attach a `Tag="Indexable"` custom property to all elements that could be searched/indexed
3. Write a small Roslyn source generator that scans all Pages/*.xaml pages, find elements that have the custom property, and generates a partial class with a list (e.g. `public static readonly (Type TargetPageType, string ResourceKey, string TargetElementId)[] SearchItems`
4. SettingsWindow.xaml.cs then just calls this list instead of the hardcoded list. It should still be easy to refresh the list if the language is changed.

This workaround should be more sustainable long-term, and we'll still see the benefits of lazy loading. Let me know what you think!

Yasinnet0:
> Hi @Yasinnet0, I've been looking into this issue and I believe we can make this work in a different way so we don't have to hardcode every searchable option in code. What I think would work much better:
> 
> 1. Remove the hardcoded `AddDeepLink()` calls
> 2. Attach a `Tag="Indexable"` custom property to all elements that could be searched/indexed
> 3. Write a small Roslyn source generator that scans all Pages/*.xaml pages, find elements that have the custom property, and generates a partial class with a list (e.g. `public static readonly (Type TargetPageType, string ResourceKey, string TargetElementId)[] SearchItems`
> 4. SettingsWindow.xaml.cs then just calls this list instead of the hardcoded list. It should still be easy to refresh the list if the language is changed.
> 
> This workaround should be more sustainable long-term, and we'll still see the benefits of lazy loading. Let me know what you think!

i think this way is good actually since this one doesn't make problems for the lazy loading i will do this one

Yasinnet0:
@unchihugo can you do a quick check if everything matches what you requested? (●'◡'●)

Yasinnet0:
Accidentally committed a local config file while dealing with the  dotnet format  checks 🤦‍♂️ Force-pushed to wipe it
and did a force push to format the merge from the actuall code changes sorry for lot of force pushes😅

unchihugo:
Looks good, I'll test it out tomorrow. Could you change the new project's LangVersion to latest instead of 12? I believe we're already using a newer C# version. I've also noticed that you're checking every \*.xaml page currently. Could you limit the check to pages in Pages/*.xaml only?

Yasinnet0:
> Looks good, I'll test it out tomorrow. Could you change the new project's LangVersion to latest instead of 12? I believe we're already using a newer C# version. I've also noticed that you're checking every *.xaml page currently. Could you limit the check to pages in Pages/*.xaml only?

all done

unchihugo:
~~Hi @Yasinnet0, looks great now! Could you just revert all the x:Name additions introduced in this PR?~~ EDIT: actually, I just looked through the code and you'll definitely need that property there for now.

FluentFlyoutWPF/Pages/AdvancedPage.xaml
```diff
@@ -26,7 +26,7 @@
                     </StackPanel>
                 </ui:CardExpander.Header>
                 <StackPanel>
-                    <ui:CardControl Icon="{ui:SymbolIcon TopSpeed24}" Margin="0,0,0,3">
+                    <ui:CardControl x:Name="AnimationDurationScaleTitleCard" Icon="{ui:SymbolIcon TopSpeed24}" Margin="0,0,0,3">
```
unchihugo: It seems like you've missed the Indexable property here

FluentFlyoutWPF/Pages/AdvancedPage.xaml
```diff
@@ -44,7 +44,7 @@
                             </ComboBox>
                         </StackPanel>
                     </ui:CardControl>
-                    <ui:CardControl Icon="{ui:SymbolIcon BezierCurveSquare12}" Margin="0,0,0,3">
+                    <ui:CardControl x:Name="AnimationEasingStyleTitleCard" Icon="{ui:SymbolIcon BezierCurveSquare12}" Margin="0,0,0,3">
```
unchihugo: and here

unchihugo (changes requested):

Yasinnet0:
> ~Hi @Yasinnet0, looks great now! Could you just revert all the x:Name additions introduced in this PR?~ EDIT: actually, I just looked through the code and you'll definitely need that property there for now.

hi yes i was about to say because its needed for the search to now exactly where to find it instead of finding it thorugh everything 

and i will fix the two missing spot right now

Yasinnet0:
all done

unchihugo (dismissed):
Perfect, works great now, thank you!

Yasinnet0:
> Perfect, works great now, thank you!

glad it works i'm always happy to help out with the project (●'◡'●)

FluentFlyoutWPF/SettingsWindow.xaml.cs
```diff
@@ -60,14 +94,53 @@ public static void ShowInstance(string? navigationPage = null)
...
+            sender.ItemsSource = matches;
+            sender.IsSuggestionListOpen = matches.Count > 0;
+        }
+    }
+
+    private void FluentWindow_PreviewMouseDown(object sender, System.Windows.Input.MouseButtonEventArgs e)
+    {
+        if (SearchBox.IsKeyboardFocusWithin && !SearchBox.IsMouseOver)
         {
-            var pageType = System.Reflection.Assembly
```
unchihugo, line 66: The following code allowed settings to instantly redirect to a specific page if requested, like the taskbar visualizer click method. Removing this breaks that functionality.

FluentFlyoutWPF/SettingsWindow.xaml.cs
```diff
@@ -113,11 +250,19 @@ await Dispatcher.InvokeAsync(async () =>
                     await Task.Delay(10);
                     RootNavigation.IsPaneOpen = wasPaneOpen;
 
-                    await Task.Delay(300);
```
unchihugo, line 116: WPF-UI refreshes the UI theme and closes the current page. Without the following code, the current page will break when changing themes.

unchihugo (changes requested):
Hi @Yasinnet0, just noticed that you refactored some parts of code that were important to the functionality of the settings window. Could you take a look and address the comments?

Yasinnet0:
> Hi @Yasinnet0, just noticed that you refactored some parts of code that were important to the functionality of the settings window. Could you take a look and address the comments?

thanks for letting me know , when i added the search deeplink i removed the original code but when i took the deeplink stuff out later i forgot to put the original back both are fixed now

unchihugo (approved):
Thank you for the contribution!

PR #965, opened 2026-07-18
Add clickable media flyout player activation

### LINKED ISSUE #934: [FEATURE REQUEST] Routing from the player to the sound source window
### Description

Hi there! It would be great to add a feature to the player that lets you jump to the media source—taking the user directly to the audio source window when clicked. This would be incredibly useful when working with multiple windows and virtual desktops, making it easier to switch to the window to add a track to a playlist or "like" a song.

### Additional context

_No response_

TeKa0:
## Summary

Adds an optional media flyout setting that lets users click the song title or player name to jump to the active media source window.

## Motivation

I added this because I personally missed a quick way to return from the media flyout to the app or browser tab that is currently playing media.

This also addresses the request from #934.

Closes #934

## Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Other

## What Changed

- Added a media flyout setting to enable opening the active player from the media panel.
- Made the song title and player name clickable when the setting is enabled.
- Added hand cursor feedback for clickable media info.
- Focuses the existing player window when possible.
- Attempts to select the matching browser tab for browser-based media sessions.
- Added English-only strings for the new setting.

## Additional Information

Related issue: #934

Tested with:

`dotnet build "FluentFlyoutWPF\FluentFlyout.csproj"`

The build completed successfully with 0 errors.

## Checklist

- [X] Code changes are manually tested and working.


FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
```diff
@@ -66,12 +68,6 @@ public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlay
             {
                 try
                 {
-                    // pre-filter processes without a main window handle
```
unchihugo, line 70: Why has this been removed?

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
```diff
@@ -120,6 +116,87 @@ public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlay
...
+
+        try
+        {
+            using var process = Process.GetProcessById(cachedInfo.ProcessId);
+            IntPtr handle = process.MainWindowHandle;
+            if (TryActivateBrowserTab(process.ProcessName, mediaTitle)) return true;
+
+            if (handle == IntPtr.Zero)
+            {
+                foreach (var candidate in Process.GetProcessesByName(process.ProcessName))
```
unchihugo, line 164: is this loop necessary? Could we return false early instead or have you noticed processes often missing the MainWindowHandle?

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
```diff
@@ -120,6 +116,87 @@ public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlay
...
+                {
+                    handle = candidate.MainWindowHandle;
+                    candidate.Dispose();
+                    if (handle != IntPtr.Zero) break;
+                }
+            }
+
+            if (handle != IntPtr.Zero)
+            {
+                if (IsIconic(handle)) ShowWindow(handle, SW_RESTORE);
```
unchihugo, line 174: Just wondering, why do we need both ShowWindow and SetForegroundWindow?

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
```diff
@@ -120,6 +116,87 @@ public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlay
...
+                return SetForegroundWindow(handle);
+            }
+
+            string? path = process.MainModule?.FileName;
+            if (!string.IsNullOrWhiteSpace(path) && !IsBrowser(System.IO.Path.GetFileNameWithoutExtension(path)))
```
unchihugo, line 179: This condition seems unnecessary, I don't think there will be a lot of cases where media plays but the process quit. I think we should just return here.

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
```diff
@@ -120,6 +116,87 @@ public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlay
...
+            {
+                Process.Start(new ProcessStartInfo(path) { UseShellExecute = true });
+                return true;
+            }
+        }
+        catch { }
```
unchihugo: Instead of leaving catch empty, let's catch all exceptions and add an error log entry.

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
```diff
@@ -120,6 +116,87 @@ public static (string, ImageSource?) GetAndCacheMediaPlayerData(string mediaPlay
...
+        if (!IsBrowser(processName) || string.IsNullOrWhiteSpace(mediaTitle)) return false;
+
+        foreach (var browserProcess in Process.GetProcessesByName(processName))
+        {
+            try
+            {
+                IntPtr handle = browserProcess.MainWindowHandle;
+                if (handle == IntPtr.Zero) continue;
+
+                var tabs = AutomationElement.FromHandle(handle).FindAll(TreeScope.Descendants,
```
unchihugo: Have you tested whether this works will all the browsers in IsBrowser? This part seems very finicky, and we might just be able to leave the TryActivateBrowserTab method out, and instead, focus the browser through the normal flow other media apps go through.

unchihugo (changes requested):
Hi @TeKa0, thanks for looking to contribute!

I think we should definitely move activation to the media player icon + name instead of the title or artist. Because of that, we can leave out the new settings and revert UI changes - turning on Show Media Player name should enable the feature.

I also left some other comments, could you please review them?

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
TeKa0: This check drops tray-only players too early. For example with Youtube Music Desktop App, when it is already hidden in the tray before FluentFlyout starts, the media panel loses the player icon and name.
The attached screenshot is from the Windows Store version and shows the issue: after starting FluentFlyout, the player name is missing, but it should show "Youtube Music Desktop App"

Image

Image

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
TeKa0: Same as with the previous comment: when the player is already in the tray before FluentFlyout starts, it has no MainWindowHandle, so returning false early means clicking it will not open/focus the app. This happens quite often for me

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
TeKa0: With only SetForegroundWindow, minimized windows do not appear on screen. The app gets focus, but the window stays minimized on the taskbar.
With only ShowWindow, clicking again on an already focused fullscreen/maximized window restore it to a smaller window size

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
TeKa0: Without this, tray-only players cannot be opened. When the player has no MainWindowHandle, there is no window to focus, so the click just returns false. This block is what lets apps like Youtube Music Desktop App reopen from the tray

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
TeKa0: sure

FluentFlyoutWPF/Classes/Utils/MediaPlayerData.cs
TeKa0: I tested this only with Chrome before. I checked Edge, firefox, opera and opera gx now too.
Firefox, opera and opera gx is problematic because sometimes FluentFlyout does not detect audio from it at all. At least on my setup. Chrome and Edge cover most browser usage so I limited this feature to those two browsers.
I think this is worth keeping because clicking the media player can open the exact tab that plays audio, not just focus any browser window

TeKa0:
> Hi @TeKa0, thanks for looking to contribute!
> 
> I think we should definitely move activation to the media player icon + name instead of the title or artist. Because of that, we can leave out the new settings and revert UI changes - turning on Show Media Player name should enable the feature.
> 
> I also left some other comments, could you please review them?

@unchihugo I replied to the other comments.

One concern about moving this only to the media player icon + name: what about a compact layout, where the media player icon and name are not visible? In that case this feature would not be available there.

unchihugo:
Thanks for your comments, i appreciate it. 

As for the compact mode, it's indeed a good issue to bring up. Perhaps we can add that in the future, but for now, I really think the UX would make much more sense if the media player is pressed. That would mean to ignore compact mode for now.

For processes without a window handle, how do they play media without a window? I'm not quite familiar with the usage flow of tray icon only media players.

TeKa0:
> As for the compact mode, it's indeed a good issue to bring up. Perhaps we can add that in the future, but for now, I really think the UX would make much more sense if the media player is pressed. That would mean to ignore compact mode for now.

Sure, done

> For processes without a window handle, how do they play media without a window? I'm not quite familiar with the usage flow of tray icon only media players.

they still have a running process and an active media session, but their main window is hidden/closed to tray, so MainWindowHandle is zero. If FluentFlyout is started after that, it still sees the media session, but the player process does not expose a normal window handle at that moment.

unchihugo (changes requested):
Hi @TeKa0, I just tested your branch, and I'm encountering hundreds of uncaught exceptions upon startup that don't throw on master:
`Exception thrown: 'System.ComponentModel.Win32Exception' in System.Diagnostics.Process.dll
`
This also hangs the UI thread for a few seconds, giving us a clue that it might potentially be related to the process search.

At line 76, I'm also getting the unhandled exception:
`System.InvalidOperationException: 'Cannot process request because the process (30644) has exited.'`
This goes on for a few times before the program fatally exits.

I'm thinking both issues are closely related to removing the code that handles filtering processes without a MainWindowHandle. 

I'd love to get this feature in the app, so I'll be here if you need help!

TeKa0:
@unchihugo Thanks for testing and pointing this out.

I adjusted the process lookup a bit, so now the `Win32Exception` count should be back to the same level as master instead of throwing many more during startup.

I could not reproduce the `InvalidOperationException` on my setup, so could you please check if it still happens after this fix?

unchihugo (dismissed):
Works great now. After testing, I believe the exceptions have now been patched. Just for consistency, do you think you could revert the changes done in FluentFlyoutWPF/Pages/MediaFlyoutPage.xaml since it's no longer needed? Thanks!

TeKa0:
> Works great now. After testing, I believe the exceptions have now been patched. Just for consistency, do you think you could revert the changes done in FluentFlyoutWPF/Pages/MediaFlyoutPage.xaml since it's no longer needed? Thanks!

Done, I reverted the MediaFlyoutPage.xaml changes

unchihugo (approved):

PR #995, opened 2026-08-04
Fix race condition causing NullReferenceException in taskbar auto-hide timer

### LINKED ISSUE #994: Taskbar widget crashes intermittently — NullReferenceException in auto-hide timer
### Description
The taskbar widget (title/visualizer) randomly disappears entirely, taking down the whole taskbar window process.

### Crash evidence (Windows Event Viewer)
Source: .NET Runtime
Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at FluentFlyout.Windows.TaskbarWindow.<UpdateUi>b__25_2(Object s, EventArgs e)
   at System.Windows.Threading.DispatcherTimer.FireTick()

App version: 2.13.1.0 (Store/MSIX build)

### Likely cause
In TaskbarWindow.xaml.cs, the auto-hide timer's Tick handler calls `_autoHideTimer.Stop()` on a field that can be nulled by a concurrent call to UpdateUi() from another thread (media-session callbacks appear to run off the UI thread). If playback resumes and clears `_autoHideTimer` right as the delayed tick fires, the handler dereferences a null field.

### Suggested fix
Capture the timer instance locally before starting it, and null-check inside the Tick handler instead of relying on the shared field:

    var localTimer = new DispatcherTimer { Interval = TimeSpan.FromMilliseconds(750) };
    localTimer.Tick += (s, e) =>
    {
        localTimer.Stop();
        if (_autoHideTimer == localTimer) _autoHideTimer = null;
        ...
    };
    _autoHideTimer = localTimer;
    localTimer.Start();

### Repro
Rapidly toggling media playback (pause/resume within ~750ms) with TaskbarWidgetAutoHide enabled.

## Testing
- Built successfully.
- Ran the app with TaskbarWidgetAutoHide enabled, rapidly toggled media playback pause/resume repeatedly — no crash observed. Previously this would intermittently trigger the NullReferenceException from issue #994.

frans15-2006:
Fixes #994

## Summary
Fixes a race condition in the taskbar auto-hide timer that could crash the app with a NullReferenceException.

## Motivation
`_autoHideTimer` was a shared field read and written from potentially different threads. If `UpdateUi()` cleared the field on one thread right as the delayed tick fired on another, the tick handler's `_autoHideTimer.Stop()` call dereferenced a null reference.

## What Changed
- Captured the timer instance locally (`localTimer`) before starting it.
- Tick handler now stops and null-checks against `localTimer` directly instead of the shared field, eliminating the race.

## Testing
- Built successfully.
- Ran the app with TaskbarWidgetAutoHide enabled, rapidly toggled media playback pause/resume repeatedly — no crash observed. Previously this would intermittently trigger the NullReferenceException from issue #994.

## Checklist

### DIFF
--- a/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
+++ b/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
@@ -691,15 +691,16 @@ on_error:
                 // Start delayed hide
                 if (_autoHideTimer == null)
                 {
-                    _autoHideTimer = new DispatcherTimer
+                    var localTimer = new DispatcherTimer
                     {
                         Interval = TimeSpan.FromMilliseconds(750)
                     };
 
-                    _autoHideTimer.Tick += (s, e) =>
+                    localTimer.Tick += (s, e) =>
                     {
-                        _autoHideTimer.Stop();
-                        _autoHideTimer = null;
+                        localTimer.Stop();
+                        if (_autoHideTimer == localTimer)
+                            _autoHideTimer = null;
 
                         if (_lastPlaybackStatus != GlobalSystemMediaTransportControlsSessionPlaybackStatus.Playing)
                         {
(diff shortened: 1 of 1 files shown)


unchihugo (approved):
Looks good!

PR #1025, opened 2026-08-14
Fix app filtering matching display names by substring

### LINKED ISSUE #1024: [BUG] App filtering matches display names by substring, so one entry filters unrelated apps
### Describe the bug

App filtering matches list entries against the app's **display name** with `Contains`, so a single entry also filters every other app whose display name happens to contain that entry as a substring. One listed app silently takes unrelated apps with it.

[`MainWindow.IsSessionAllowed()`](https://github.com/unchihugo/FluentFlyout/blob/master/FluentFlyoutWPF/MainWindow.xaml.cs#L265-L291) (identical in both branches of the mode check):

```csharp
BlockedApps.Any(b =>
    appName.Contains(b, StringComparison.OrdinalIgnoreCase) ||   // <-- here
    appId.Contains(b, StringComparison.OrdinalIgnoreCase))
```

`appName` is the user-facing display name from `MediaPlayerData.GetAndCacheMediaPlayerData()`, and the substring test is applied to it.

**What I observed.** I have two apps that both publish SMTC sessions, `Amazon Music` and `AmazonMusic SMTC Bridge`. With only `Amazon Music` on the blocklist, both were hidden from the flyout and the taskbar widget — the second app was never listed, and blocking it was not what I asked for. The name pair is not special; the same thing happens for any pair like `Spotify` / `Spotify Widget` or `Google Chrome` / `Google Chrome Beta`.

This is not limited to blocklist mode. The allowlist branch has the same predicate, where it shows up as unlisted apps being *allowed*.

**Minimal reproduction (no second app needed).** Entries typed into the filtering page are normalized to a full session name by [`NormalizeAppName()`](https://github.com/unchihugo/FluentFlyout/blob/master/FluentFlyoutWPF/Pages/AppFilteringPage.xaml.cs#L39-L56), so the substring behaviour has to be reached by writing the entry directly:

1. Start any media app — say Spotify, shown as `Spotify` in Settings > App filtering.
2. Enable app filtering in blocklist mode, then close FluentFlyout.
3. In `%APPDATA%\FluentFlyout\settings.xml`, put a partial name in `BlockedApps`:
   ```xml
   <BlockedApps>
     <string>Spot</string>
   </BlockedApps>
   ```
4. Start FluentFlyout and play something in Spotify.

Spotify is filtered out, even though no app is named `Spot`. Any app whose display name contains `Spot` would be filtered the same way.

**Why the substring test looks unintended.** Only two things write these entries, and both store a complete app name:

- the `AppFilteringPage` ComboBox, which stores the `GetAndCacheMediaPlayerData()` return value verbatim
- the manual text box, which runs `NormalizeAppName()` to resolve what was typed to a real session name first

So the stored value is a whole name by construction, and matching it with `Contains` is the part that disagrees.

**Why it surfaced now.** Before v2.14.0 there was no `ResolveViaAppsFolder()`, so `appName` fell back to the raw AUMID (`AmazonMusicSmtc_f91r0mcej9kn8!App`), which happens not to contain `Amazon Music`. Display names became resolvable in `1f9ea47`, first shipped in v2.14.0, and the collision became reachable. The faulty comparison itself is older.

I have a fix ready and will open a PR against this issue: match the display name with `Equals`, and keep `Contains` on the session id so existing entries that were registered as a package or publisher fragment still work.

### Logs

App filtering does not write any log entries — `IsSessionAllowed()` runs on every session evaluation and logs nothing — so there is no log line corresponding to a session being hidden. Startup log from a session where the bug was present is attached below; the `settings.xml` state at the time was `AppFilteringEnabled=true`, `AppFilteringMode=0`, `BlockedApps=[Amazon Music]`.

```
2026-08-14 13:07:57.4787|INFO|FluentFlyoutWPF.MainWindow|Starting FluentFlyout MainWindow
2026-08-14 13:07:57.6278|INFO|FluentFlyout.Classes.Settings.SettingsManager|Settings successfully restored
2026-08-14 13:07:57.8279|INFO|FluentFlyout.Classes.LocalizationManager|Applying localization for language: ja-JP
2026-08-14 13:07:57.8279|INFO|FluentFlyout.Classes.LocalizationManager|Applied flow direction: LeftToRight
2026-08-14 13:07:57.9242|INFO|FluentFlyoutWPF.MainWindow|Current version: v2.14.0
2026-08-14 13:07:58.3861|INFO|FluentFlyout.Classes.LicenseManager|LicenseManager: Initializing
2026-08-14 13:07:58.4008|INFO|FluentFlyoutWPF.MainWindow|License synced on startup - Store: False, Premium: True
2026-08-14 13:07:58.4008|INFO|FluentFlyout.Classes.Settings.SettingsManager|Settings successfully saved to C:\Users\...\AppData\Roaming\FluentFlyout\settings.xml
2026-08-14 13:07:58.4823|INFO|FluentFlyoutWPF.Classes.Services.UpdateCheckerService|Update check complete. Current: v2.14.0, Newest: v2.14.0, Update available: False
2026-08-14 13:07:58.8080|INFO|FluentFlyoutWPF.Classes.AudioDeviceMonitor|Audio device monitoring initialized
2026-08-14 14:44:48.9471|INFO|FluentFlyoutWPF.Classes.AudioDeviceMonitor|Default audio output device changed
2026-08-14 14:44:48.9471|INFO|FluentFlyoutWPF.ViewModels.VolumeMixerViewModel|Default render device changed, reattaching volume mixer
2026-08-14 14:51:05.4609|INFO|FluentFlyout.Classes.Settings.SettingsManager|Settings successfully saved to C:\Users\...\AppData\Roaming\FluentFlyout\settings.xml
```

### Screenshots & additional context

Display names as resolved through `shell:AppsFolder` on this machine, which is the same path `ResolveViaAppsFolder()` uses:

```
Amazon Music             ==>  AmazonMobileLLC.AmazonMusic_kc6t79cpj4tp0!AmazonMobileLLC.AmazonMusic
AmazonMusic SMTC Bridge  ==>  AmazonMusicSmtc_f91r0mcej9kn8!App
```

Note that the two AUMIDs share no substring with each other, so matching on the session id alone would never have confused them — the collision is introduced by the display name comparison.

Behaviour of the current predicate versus an exact display-name comparison, for the entries above:

| Entry | Session | Current | Exact display name |
|---|---|---|---|
| `AmazonMusic SMTC` | `AmazonMusic SMTC Bridge` | filtered | not filtered |
| `Amazon Music` | `Amazon Music` | filtered | filtered |
| `Amazon Music` | `AmazonMusic SMTC Bridge` | not filtered | not filtered |
| `AmazonMobileLLC` | `Amazon Music` | filtered | filtered (id match) |
| `AmazonMobileLLC` | `AmazonMusic SMTC Bridge` | not filtered | not filtered |
| `` (empty entry) | any | filtered | not filtered |

The last row is a second, smaller consequence of the same line: an empty entry is a substring of everything, so in blocklist mode it hides every session and in allowlist mode it allows every session.

### FluentFlyout Version

v2.14.0

### Windows Version

Windows 11 Home 10.0.26200

### Architecture

x64

### Checklist

Fuku856:
## Summary

App filtering compared list entries against the session's **display name** with `Contains`, so a single entry also filtered every other app whose display name contained it as a substring. This makes the display name comparison exact, and leaves the session id on substring matching.

## Motivation

Closes #1024

`IsSessionAllowed()` ran the same predicate in both modes:

```csharp
appName.Contains(b, StringComparison.OrdinalIgnoreCase) ||
appId.Contains(b, StringComparison.OrdinalIgnoreCase)
```

`appName` is the user-facing name from `GetAndCacheMediaPlayerData()`, so blocking one app silently blocked unrelated apps whose name happened to contain it — I hit this with `Amazon Music` on the blocklist also hiding `AmazonMusic SMTC Bridge`, but the same holds for any pair like `Spotify` / `Spotify Widget`. Allowlist mode has the mirrored problem, letting unlisted apps through.

Matching the display name in full is what the rest of the feature already assumes. Only two things write these entries, and both store a complete app name:

- the `AppFilteringPage` ComboBox stores the `GetAndCacheMediaPlayerData()` return value verbatim
- the manual text box runs `NormalizeAppName()`, which resolves what was typed to a real session name first

So the stored value is a whole name by construction; the `Contains` on the matching side was the half that disagreed.

The session id deliberately keeps substring matching. Entries added before the display name became resolvable can hold a fragment like `chrome` or `spotify` that only lines up with the AUMID, and that is also the way an entry can target a whole package or publisher rather than one display name. Tightening both sides would be more uniform but would silently break those existing settings, so it is left alone.

Blank entries are now skipped as well — an empty string is a substring of everything, so it previously hid every session in blocklist mode and allowed every session in allowlist mode.

## Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Other

## What Changed

- `MainWindow.IsSessionAllowed()` — the predicate duplicated across the blocklist and allowlist branches is extracted into a new `MatchesFilterEntry()` helper, so both modes stay in step.
- `MatchesFilterEntry()` compares the display name with `Equals` instead of `Contains`, keeps `Contains` for the session id, and returns false for null/whitespace entries.
- One file touched, `FluentFlyoutWPF/MainWindow.xaml.cs`. No settings, UI or localization changes, so no resource dictionary updates are needed.

## Additional Information

Behaviour for the two apps I reproduced with (`Amazon Music` → `AmazonMobileLLC.AmazonMusic_kc6t79cpj4tp0!…`, `AmazonMusic SMTC Bridge` → `AmazonMusicSmtc_f91r0mcej9kn8!App`):

| Blocklist entry | Session | Before | After |
|---|---|---|---|
| `AmazonMusic SMTC` | `AmazonMusic SMTC Bridge` | blocked | **allowed** — partial display name, and the AUMID has no space so the id does not match either |
| `Amazon Music` | `Amazon Music` | blocked | blocked (exact match, unchanged) |
| `Amazon Music` | `AmazonMusic SMTC Bridge` | allowed | allowed |
| `AmazonMobileLLC` | `Amazon Music` | blocked | blocked — the id substring path still works |
| `AmazonMobileLLC` | `AmazonMusic SMTC Bridge` | allowed | allowed |
| `` (blank) | any | blocked | **allowed** |

I kept the diff confined to `IsSessionAllowed()` and its new helper, so if #1023 lands first this should only need the `BlockedApps` / `AllowedApps` / `AppFilteringMode` property renames applied to the two call sites.

## Checklist

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -272,22 +272,30 @@ public partial class MainWindow : MicaWindow
 
         if (SettingsManager.Current.AppFilteringMode == 0) // Blacklist mode
         {
-            if (SettingsManager.Current.BlockedApps != null && SettingsManager.Current.BlockedApps.Any(b =>
-                    appName.Contains(b, StringComparison.OrdinalIgnoreCase) ||
-                    appId.Contains(b, StringComparison.OrdinalIgnoreCase)))
+            if (SettingsManager.Current.BlockedApps != null &&
+                SettingsManager.Current.BlockedApps.Any(b => MatchesFilterEntry(b, appName, appId)))
                 return false;
 
             return true;
         }
         else // Whitelist mode
         {
-            if (SettingsManager.Current.AllowedApps != null && SettingsManager.Current.AllowedApps.Any(a =>
-                    appName.Contains(a, StringComparison.OrdinalIgnoreCase) ||
-                    appId.Contains(a, StringComparison.OrdinalIgnoreCase)))
+            if (SettingsManager.Current.AllowedApps != null &&
+                SettingsManager.Current.AllowedApps.Any(a => MatchesFilterEntry(a, appName, appId)))
                 return true;
 
             return false;
         }
+    }
+
+    // display names are matched in full: entries always hold a whole app name, and a substring match let one entry
+    // swallow others (blocking "Amazon Music" also blocked "Amazon Music SMTC Bridge"). the session id keeps substring
+    // matching so an entry can still target a whole package or publisher
+    private static bool MatchesFilterEntry(string entry, string appName, string appId)
+    {
+        if (string.IsNullOrWhiteSpace(entry)) return false; // an empty entry would match every session
+
+        return appName.Equals(entry, StringComparison.OrdinalIgnoreCase) || appId.Contains(entry, StringComparison.OrdinalIgnoreCase);
     }
 
     public MediaSession? GetActiveMediaSession()


unchihugo (commented):
Hi, thanks for the PR fixing the oversight. Before we merge, I was wondering why you decided to still keep .Contains for appId specifically?

Fuku856:
Thanks for taking a look!

Short version: keeping `Contains` on `appId` is what keeps this from being a
breaking change, and one of the existing normalisation steps only works because
of it.

**1. It's the only thing that can match an entry after `NormalizeAppName()`
strips `.exe`.**

`NormalizeAppName()` removes a trailing `.exe`, then tries to resolve what's
left against the *live* sessions and returns `match ?? app`. If that app isn't
currently playing there's nothing to resolve against, so the stripped string is
what gets saved — type `chrome.exe` while Chrome is closed and the stored entry
is `chrome`.

That entry then has to match a session whose id is `chrome.exe`. `Equals`
wouldn't, `Contains` does. So stripping the extension on the way in and
requiring an exact id match on the way out can't both be right; the strip only
makes sense if the id side stays a substring match.

**2. AUMIDs aren't names, and an entry can legitimately be a fragment of one.**

Substring matching is how an entry targets a package or publisher rather than a
single display name — e.g. `AmazonMobileLLC` catching
`AmazonMobileLLC.AmazonMusic_kc6t79cpj4tp0!AmazonMobileLLC.AmazonMusic`.
`MediaPlayerData` already treats ids as structured this way: it splits on `.`,
strips `com`/`github`/`exe`, and special-cases `!MSEdge`.

**3. The collisions are on the display name side, not the id side.**

Display names are human-facing, so one containing another is normal (`Spotify` /
`Spotify Widget`, `Google Chrome` / `Google Chrome Beta`). AUMIDs and exe names
are namespaced and collide far less often. Making the name exact removes the bug
class while leaving the id escape hatch intact.

To be upfront about the limit: the id side isn't collision-proof either — two
packages from the same publisher can still catch each other. I left it alone
deliberately, since tightening it would silently stop existing settings from
working, which felt like more than a bug fix should do. If you'd prefer both
sides exact for consistency I'm happy to change it, though it'd probably want a
release note, and `NormalizeAppName()`'s `.exe` strip would need revisiting at
the same time.

unchihugo (approved):
Sounds good! Thanks for the insight.

PR #1030, opened 2026-08-15
Fix: Unexpected main window popup on startup when auto-launching

oldeightLB:
It fixes an issue where the main window unexpectedly pops up on startup when auto-launching in self-compiled versions.
Some Markdown file changes may have been introduced accidentally while using OpenCode. Please ignore them if they are unrelated to this fix.


unchihugo:
There are a lot of unnecessary changes in the PR and the PR template has not been followed. 

Furthermore, I do not understand the necessity of this system, as the main window only opens upon the second launch of FluentFlyout (given that it is a Singleton program), which appears to make this redundant or unnecessary.

I will have to close this for now, sorry!

PR #1060, opened 2026-08-23
Fix media flyout control centering

### LINKED ISSUE #536: [FEATURE REQUEST] Center Title and Artist should also center the button container

- [x] I have read the [FAQ](https://fluentflyout.com/faq/) and there is no answer to my request

### Description

Title and Artist are centered, but the button container below is left-aligned. It looks slightly better if the player app information is on, but when that gets hidden, the result is strongly unbalanced:

Image

The button container should be centered dynamically based on its width and the buttons that are enabled / visible

If hiding even more buttons, the whole container gets smaller (it shouldn't), without the issue being solved:

Image

### Additional context

_No response_

lgc2333:
Fixes #536

Center media controls with title/artist when the player name is hidden.

image

image

image

compact mode works fine:

image

image


unchihugo (changes requested):
Hi @lgc2333, looks great so far! I've noticed I believe the alignment should default to left instead of right. Also, instead of always calculating the width, do you think it's possible to let the stack panel automatically fully span the parent horizontally so we just have to flip alignment left to center?

lgc2333:
implemented. adapted some code to keep the existing 2px left margin at other scenarios.

unchihugo (commented):
Thanks for the improvements! I've tested myself and I do see a slight margin error if the media player display is enabled:

Image

As you can observe, there's a slight cutoff visible when hovering over the button (left side has rounded corners - right side doesn't).

Other than that, good work so far!

lgc2333:
fixed. the left 2px margin only applies at compact mode.

the 2px left margin is never actually applied at non-compact mode originally maybe.

lgc2333:
is the layout right at compact mode? i personally think the close button's left margin is too big. it matches the original layout tho

image

image

unchihugo:
> is the layout right at compact mode? i personally think the close button's left margin is too big. it matches the original layout tho
> 
> image image

Looks good to me, I'll test it on my machine soon. 

In the future, we'll have to prioritize moving away from manually updating UI in code every time the flyout is displayed and switch to MVVM.

unchihugo (changes requested):
Looks good now! Thank you for the contribution :)

unchihugo (approved):

PR #1061, opened 2026-08-23
Add taskbar widget cover position setting

lgc2333:
Adds a cover position setting for the taskbar widget: Default, Left, and Right.

Default keeps the cover on the left unless the widget position is bottom-right. When the cover is on the right, the title and artist text are right-aligned.

image

image

AI tools were used.


lgc2333:
close due to low possibility merging. turning into FR issue: #1116

PR #1068, opened 2026-08-25
Fix visualizer bar antialiasing asymmetry

### LINKED ISSUE #911: [BUG] Visualiser bars aren't perfectly symmetrical
### Describe the bug

I might be imagining things but to me it's clear that the visualiser bars are not symmetrical. The right side of every bar seems cut off and more straight than the others.

Image

### To Reproduce

_No response_

It just looks like it's had about 1 or 2 pixels cut off the right side. This is noticeable to me even not zoomed in.

Image

### FluentFlyout Version

v2.12.2

### Windows Version

Windows 11 25H2

### Architecture

x64

### Checklist

vermilion10:
## Summary

Fixes an antialiasing asymmetry in the taskbar visualizer: the left edge of
every bar was antialiased while the right edge was not, which made the bars
look clipped on the right.

## Motivation

Closes #911

`RasterizeBar` sampled each pixel at its integer top-left coordinate, while the
bar geometry spans the continuous range `[barX, barX + barWidth]`. This biased
the whole sample grid half a pixel to the left:

- the leftmost column sampled exactly on the shape edge, so the corner SDF gave
  it partial coverage (`alpha ~= 0.5`), producing a soft edge;
- the rightmost column sampled a full unit inside the shape, so it was always
  fully opaque, producing a hard edge.

The same bias applied vertically, giving feathered top corners and hard bottom
corners.

This also explains why the severity looked inconsistent between machines in the
issue thread. The visualizer bitmap is a 3x supersample (228x96) scaled down
into a ~76x32 image, so the error is a sub-pixel edge-softness difference whose
visibility depends on the display scaling factor. That matches the report of
the right side looking "1 or 2 pixels cut off": the bar is not narrower, it is
just missing the feathering the left side gets.

## Type of Change

- [ ] Feature
- [x] Bug fix
- [ ] Other

## What Changed

- `RasterizeBar` now samples at pixel centers (`x + 0.5f`, `y + 0.5f`) instead of
  integer pixel coordinates, so the outermost samples sit half a pixel inside
  each edge and both sides receive identical coverage.
- Loop bounds, buffer indexing, and the layout/radius helpers are unchanged.

Net diff is 10 insertions / 7 deletions in one file.

## Additional Information

I measured per-column alpha coverage of a rendered bar and compared each column
against its mirror. Horizontal mirror error before/after, at bar height 40:

| Bar count | Centered | Before | After |
|-----------|----------|--------|-------|
| 10        | yes      | 4.200  | 0.000 |
| 10        | no       | 2.100  | 0.000 |
| 11        | yes      | 4.497  | 0.000 |
| 11        | no       | 2.249  | 0.000 |
| 12        | yes      | 3.600  | 0.000 |
| 12        | no       | 1.800  | 0.000 |

visualizer-before-after

Vertical symmetry also reaches 0.000 for centered bars. It stays non-zero for
non-centered bars, which is expected since those are intentionally flat-bottomed.

`dotnet build` and `dotnet format --verify-no-changes` both pass, with the same
48 pre-existing warnings as master (this change adds none).

Scoping note: `ComputeLayout` subtracts 1 from the available width and truncates
when centering, which can leave the whole bar strip 1px off-centre for some bar
counts (11 bars gives a 1px left margin and a 2px right margin). That is a
separate whole-strip offset rather than the per-bar asymmetry reported here, so
I left it out of this PR. Happy to address it separately if you'd like.

## Checklist

### DIFF
--- a/FluentFlyoutWPF/Classes/Visualizer.cs
+++ b/FluentFlyoutWPF/Classes/Visualizer.cs
@@ -19,14 +19,9 @@ namespace FluentFlyoutWPF.Classes
 
         public static int BarCount = 10;
 
-        // The bitmap is supersampled and then minified by WPF to its on-screen size.
-        // Keep this at 2: at an exact 2:1 minification the default Linear filter samples
-        // the midpoint between two source pixel centers, so its weights land at exactly
-        // 50/50 and the tap becomes a true 2x2 box average. Other factors (3, 4) put the
-        // sample on a single source pixel center or an off-centre pair, which throws most
-        // of the supersampled detail away.
+        // Keep at 2: only an exact 2:1 minification turns WPF's default Linear
+        // filter into a true box average. Other factors discard the supersampling.
         private const int Supersample = 2;
-
         private readonly int ImageWidth = 76 * Supersample;
         private readonly int ImageHeight = 32 * Supersample;
         private readonly int BarSpacing = 2 * Supersample;
@@ -485,7 +480,7 @@ namespace FluentFlyoutWPF.Classes
             byte r = brush.Color.R;
 
             bool centeredBars = SettingsManager.Current.TaskbarVisualizerCenteredBars;
-            int barBaseline = SettingsManager.Current.TaskbarVisualizerBaseline ? Supersample : 0;
+            int barBaseline = SettingsManager.Current.TaskbarVisualizerBaseline ? Supersample * 2 : 0;
 
             int centerY = ImageHeight / 2;
 
--- a/FluentFlyoutWPF/Classes/Visualizer.cs
+++ b/FluentFlyoutWPF/Classes/Visualizer.cs
@@ -592,6 +592,7 @@ namespace FluentFlyoutWPF.Classes
             for (int y = barY; y < barEndY && y < ImageHeight && y >= 0; y++)
             {
                 int row = y * stride;
+                float py = y + 0.5f;
 
                 for (int x = barX; x < barX + barWidth && x < ImageWidth; x++)
                 {
@@ -599,33 +600,35 @@ namespace FluentFlyoutWPF.Classes
                     if (index + 3 >= buffer.Length)
                         continue;
 
+                    float px = x + 0.5f;
+
                     // CENTER
-                    if (x >= innerLeft && x <= innerRight)
+                    if (px >= innerLeft && px <= innerRight)
                     {
                         WritePixel(buffer, index, b, g, r, 255);
                         continue;
                     }
 
                     // SIDES
-                    if (y >= innerTop && y <= innerBottom)
+                    if (py >= innerTop && py <= innerBottom)
                     {
                         WritePixel(buffer, index, b, g, r, 255);
                         continue;
                     }
 
                     // FLAT BOTTOM
-                    if (!centeredBars && y >= innerBottom)
+                    if (!centeredBars && py >= innerBottom)
                     {
                         WritePixel(buffer, index, b, g, r, 255);
                         continue;
                     }
 
                     // CORNERS
-                    float cx = x < innerLeft ? innerLeft : (x > innerRight ? innerRight : x);
-                    float cy = y < innerTop ? innerTop : (y > innerBottom ? innerBottom : y);
+                    float cx = px < innerLeft ? innerLeft : (px > innerRight ? innerRight : px);
+                    float cy = py < innerTop ? innerTop : (py > innerBottom ? innerBottom : py);
 
-                    float dx = x - cx;
-                    float dy = y - cy;
+                    float dx = px - cx;
+                    float dy = py - cy;
 
                     float distSq = dx * dx + dy * dy;
                     float sdf = (distSq - radiusSq) / (2f * radius);
--- a/FluentFlyoutWPF/Classes/Visualizer.cs
+++ b/FluentFlyoutWPF/Classes/Visualizer.cs
@@ -18,9 +18,18 @@ namespace FluentFlyoutWPF.Classes
         private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
 
         public static int BarCount = 10;
-        private readonly int ImageWidth = 76 * 3;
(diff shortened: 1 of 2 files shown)


unchihugo:
Hi @vermilion10, thank you for working on this! Looks great on your example images, but on my end, it seems like AA is still not quite right at 100% DPI:
image
As you can see, bars are sometimes still more harshly rounded. I believe the anti aliasing method is still not complete. I think we'll need a bit more investigation into this before we merge!

vermilion10:
@unchihugo thanks for testing that! you're right, and i think i found the reason. ti isn't in the rasterizer, it's in how WPF scales the bitmap down

the bitmap is 228x96 (76\*3 x 32\*3), and the `Image` in `TaskbarVisualizerControl.xaml` resolves to exactly 76x32 at 100% DPI, so its an exact 3:1 minification. there's no `BitmapScalingMode` set on it, so it gets the default `Linear` filter. At a 3:1 ratio, destination pixel `j` samples source coordiante `(j+0.5)*3 = 3j+1.5`, which lands exactly on the center of source pixel `3j+1`, so the bilinear tap collapses onto that one sample. WPF is effectively point sampling: 8 of every 9 supersampled pixels get thrown away, and the 3x supersampling does nothing at 100% DPI

which of the 9 survives depends on where the bar sits relative to the 3px grid. With 10 bars, `barWidth` is 17 and spacing is 6, so `barX` advances by 23 each time and cycles through all three phases. Bars landing on phase 0 sample straight through the flat middle of the bar and come out compeltely square, which is what the third bar in your screenshot is doing. that's the "sometimes more harshly rounded" you were seeing

i confirmed it by rendering the real bitmap through an actual WPF `Image` at the real scale and capturing it with `RenderTargetBitmap`:

| DPI | default (linear) | HighQuality (Fant) |
|---|---|---|
| 100% | 100 AA px, 50 alpha level | 713 AA px, 125 levels |
| 125% | 330, 74 | 756, 130 |
| 150% | 440, 103 | 950, 142 |
| 200% | 886, 122 | 1581, 156 |

100% DPI is the worst case by a wide margin, since it's the only ratio where the bilinear taps degenerate like that. at 125% and 150% the mapping is fractional so it does actualy blend, which is probably why it only stood out to you at 100%

so the fix is one line:

```xml
<Image x:Name="VisualizerContainer" Margin="4,0" RenderOptions.BitmapScalingMode="HighQuality"/>
```

image

also, two things worth flagging. Fant is a little softer than the current hard-snapped look. At taskbar size it reads as smooth rather than blurry to me, but it is visual change, so say the word if you'd rather approach it differently. I also tried snapping the bar geometry to multiples of 3 so the straight edges stay on display-pixel boundaries, but it didnt come out any sharper (Fant's kernel is wider than a box filter regardless), so i left that out to keep the change small

the pixel-center fix in the first commit is still needed, since it's what makes the source bitmap symetric to begin with. it just wasn't sufficient on its own

unchihugo:
Thanks for the analysis! This helps out a lot. Perhaps we should move away from the 3:1 scaling, and move to something closer like 2:1? It is indeed quite blurry on 100% DPI with HighQuality AA on my end.

unchihugo (changes requested):
> Perhaps we should move away from the 3:1 scaling, and move to something closer like 2:1? It is indeed quite blurry on 100% DPI with HighQuality AA on my end.

I've investigated a bit into this: In `Visualizer.cs`, if we just adjust the readonly ints `ImageWidth`, `ImageHeight` and `BarSpacing` to `* 2` instead of `* 3`, and change `6f` to `4f` in `Visualizer.GetCornerRadius`, we should get a much better ratio for all common DPI scales.

Once that's done, I'm happy to merge the PR!

unchihugo (approved):

vermilion10:
that's exactly what cc80775 does, so this should be good to go. Measured against the analytic coverage of the intended shapes, RMSE at 100% DPI drops from 0.0984 to 0.0079, and it holds up at 6, 13 and 20 bars too

one thing beyond your list: `barBaseline` in `DrawBars` was also in supersampled units (that literal `4`), so leaving it alone would have made the resting line 50% thicker on screen. i've tied it to the same constant, which puts it at 1.0 display px against the 1.33 it was before. Rounding the other way gives 2.0, so say the word if you'd prefer the chunkier line and I'll flip it

also pulled the factor into a single `Supersample` const so the width, height, spacing, radius and baseline all derive from it, instead of the ratio being implied in five separaet places

FluentFlyoutWPF/Classes/Visualizer.cs
```diff
@@ -476,7 +485,7 @@ private unsafe void DrawBars(int stride, Span<byte> buffer)
             byte r = brush.Color.R;
 
             bool centeredBars = SettingsManager.Current.TaskbarVisualizerCenteredBars;
-            int barBaseline = SettingsManager.Current.TaskbarVisualizerBaseline ? 4 : 0;
+            int barBaseline = SettingsManager.Current.TaskbarVisualizerBaseline ? Supersample : 0;
```
unchihugo: Could we do `Supersample * 2` here?

FluentFlyoutWPF/Classes/Visualizer.cs
```diff
@@ -18,9 +18,18 @@ public class Visualizer : IDisposable
         private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
 
         public static int BarCount = 10;
-        private readonly int ImageWidth = 76 * 3;
-        private readonly int ImageHeight = 32 * 3;
-        private readonly int BarSpacing = 2 * 3;
+
+        // The bitmap is supersampled and then minified by WPF to its on-screen size.
```
unchihugo: I think the long comments here are unnecessary as the variable itself is pretty self-explanatory, and it definitely reads like the thought process of AI. You could choose to remove them or shorten them down!

unchihugo (changes requested):
Perfect! Could you address two small things I've added as comments before we merge?

unchihugo (approved):
Perfect. Thank you for your contribution!

PR #1062, opened 2026-08-23
Customize Next Up flyout

lgc2333:
Adds independent Next Up flyout controls for position, title/artist alignment, and the Up next label.

Closes #457

image

samples:

image

image

AI tools were used.

PR #1058, opened 2026-08-23
Add volume flyout sync toggle

lgc2333:
## Summary

Adds a volume setting above the volume flyout stay-duration control for linking manual media flyout triggers with the volume flyout.

Taskbar widget and tray "show media flyout" entries now go through a wrapper that:
- shows media first, then volume, when opening
- starts the volume hide animation before the media hide animation on toggle-close, without waiting between them

These popups are visually syncing when open btw

Media-key and volume-key paths keep their existing direct calls. Auto stay-duration behavior is still timer-driven, so matching media/volume durations naturally collapse together without extra close synchronization.

image

## Motivation

closes #1050

## Type of Change

- [x] Feature

## Checklist
- [x] Code changes are manually tested and working.
- [x] Formatting and naming are consistent with the project.
- [x] Self-review of changes is done.
- [x] AI tools were used (if yes, I reviewed and fully understand the changes myself).

PR #1070, opened 2026-08-25
Fix bottom-center flyout offset

lgc2333:
## Summary

Fix the extra bottom-center offset for media flyouts when FluentFlyout's custom volume flyout is stacked above them.

## Motivation

The 80px offset is only needed when the media flyout has to leave room for the native Windows volume OSD. The previous implementation put that decision in the shared flyout positioning path, so other flyouts such as Next Up could inherit media-only spacing.

## Type of Change

- [x] Bug fix

## What Changed

- Added an explicit `reserveNativeVolumeOsdSpace` positioning flag.
- Enabled that flag only for the media flyout and for the custom volume flyout when it calculates the media flyout reference position.
- Left other flyouts, including Next Up, on the normal 16px bottom-center margin.

## Screenshots

10323b9f-ad13-45b1-a175-e990eb9fe47e

6c6ae9a3-14e9-4061-b8f8-6cb73caf591c

works fine with auto hide taskbar disabled:

image

## Additional Information

Validated with `dotnet build FluentFlyoutWPF\FluentFlyout.csproj -p:UseAppHost=false`.

## Checklist

### DIFF
--- a/FluentFlyoutWPF/MainWindow.xaml.cs
+++ b/FluentFlyoutWPF/MainWindow.xaml.cs
@@ -398,6 +398,11 @@ public partial class MainWindow : MicaWindow
     /// Computes the final resting position (left, top) for a window based on the current
     /// position setting and the selected monitor's work area.
     /// </summary>
+    private static double GetBottomCenterFlyoutBottomMargin()
+    {
+        return SettingsManager.Current.VolumeControlEnabled && SettingsManager.Current.VolumeControlAboveMediaFlyout ? 16 : 80;
+    }
+
     private (double left, double top) GetFinalPosition(Rect windowRect, Rect workArea)
     {
         int position = SettingsManager.Current.Position;
@@ -410,7 +415,7 @@ public partial class MainWindow : MicaWindow
         double top = position switch
         {
             0 or 2 => workArea.Top + workArea.Height - windowRect.Height - 16,
-            1 => workArea.Top + workArea.Height - windowRect.Height - 80,
+            1 => workArea.Top + workArea.Height - windowRect.Height - GetBottomCenterFlyoutBottomMargin(),
             _ => workArea.Top + 16
         };
         return (left, top);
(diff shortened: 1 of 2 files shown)


lgc2333:
~~next-up flyout seems to be fixed bottom padding. i'll fix this in #1062~~ my mistake it's affected by this change

lgc2333:
now any bottom-center flyouts are affected. (next-up, capslock indicator, ...)

image

image

unchihugo (changes requested):
Hi @lgc2333, thanks for working on this! Instead of passing through a new parameter for the methods, why don't we just determine the positioning in the method itself? So if the user's positioning is bottom center, the code adapts instead of having to pass through a new `reserveNativeVolumeOsdSpace` boolean?

lgc2333:
(my eng not good so i use mother lang) 因为这边我们只抬高媒体浮窗，所以我觉得入参是更干净的处理方式。其实我的第一版实现（第一个commit）就是你说的意思（你可以看一下），但是这个方案有个问题：当音量浮窗没启用以及它没开启显示在下方时，其他浮窗也会收到影响（nextup,capslock），抬高下方margin

lgc2333:
我希望其他浮窗能始终贴底不抬高，因为它们不需要避让windows原生音量浮窗

unchihugo:
I see what you mean now - I'll take a look later today!

FluentFlyoutWPF/MainWindow.xaml.cs
```diff
@@ -477,12 +485,13 @@ public void OpenAnimation(MicaWindow window, bool alwaysBottom = false, MonitorI
...
             {
+                double bottomMargin = GetBottomCenterFlyoutBottomMargin(reserveNativeVolumeOsdSpace);
                 window_left = workArea.Left + workArea.Width / 2 - windowRect.Width / 2;
-                moveAnimation.To = workArea.Top + workArea.Height - windowRect.Height - 80;
+                moveAnimation.To = workArea.Top + workArea.Height - windowRect.Height - bottomMargin;
                 if (SettingsManager.Current.FlyoutAnimationSpeed == 0)
                     moveAnimation.From = moveAnimation.To;
                 else
-                    moveAnimation.From = workArea.Top + workArea.Height - windowRect.Height - 60;
+                    moveAnimation.From = moveAnimation.To + 20;
```
unchihugo: Nit: Perhaps we should create a variable for `moveAnimation.To` and `moveAnimation.From` to read from instead of `moveAnimation.From = moveAnimation.To + 20;`?

unchihugo (commented):
Looks good, thank you for your help! If you're ready to merge, just give the go-ahead.

lgc2333:
okay. also take a look if u have time #1058 #1061 #1062
thank u for your patience uwu

unchihugo (approved):

PR #22, opened 2024-11-06
Add theme switcher

LiAuTraver:
Add light, dark, and follow the system theme option. Default to follow the System theme, user can change by toggling the settings(see pic below), and the theme will be remembered.
Worked fine with `Auto Dark Mode`.

Related feature request:
https://github.com/unchihugo/FluentFlyout/issues/6#issue-2629787980

image

*p.s. Mainly not a c sharp dev, but love your app and hope this can help the app get better!* 😊

PR #1109, opened 2026-09-06
Fix 22 bugs: SMTC crashes, mixer leaks/races, seekbar freeze, settings leak, UIA stalls, OSD, startup

frans15-2006:
## Summary
Implements 22 previously unfiled bugs found via code-level review (not yet reported as GitHub issues). One commit per fix (24 commits — bug #22 ships as four scoped commits). No open issues are referenced; every change is scoped to a specific root cause found in the code.

### Media flyout (SMTC)
- **Null session crash** — `UpdateUI` guard is now `mediaSession?.ControlSession == null`, so the "No media playing" reset runs instead of NRE-ing when app filtering removes the last session.
- **Dead-session COM crashes** — added `TryGetPlaybackInfo` / `TryGetTimelineProperties` helpers and converted every raw `GetPlaybackInfo()` / `GetTimelineProperties()` call site; a session dying mid-event now logs instead of killing the process.
- **Seekbar freeze** — the slider now captures the mouse for the drag and ends it on `LostMouseCapture`, so releasing outside the slider can no longer leave `_isDragging` stuck true.
- **Back button** — gated on `IsPreviousEnabled` (was reusing `IsNextEnabled`), matching the taskbar widget.
- **Next Up popup** — the auto-close continuation is aborted via the `Closed` event when the window is closed early, instead of animating a dead window inside `async void`.
- **Packaged players** — "Open player" launches `shell:AppsFolder\<AUMID>` when the cached process id is -1, instead of throwing inside `Process.GetProcessById`.

### Volume mixer / audio
- **MMDevice leaks** — every re-attach path now disposes the superseded `MMDevice`.
- **Dead-session writes** — volume/mute writes wrap the WASAPI calls in try/catch and keep the stale value instead of throwing `COMException` out of a WPF binding setter.
- **Zero-volume ↔ mute decoupling** — mute no longer auto-triggers at 0 volume; it stays an independent toggle (matches native mixer behavior).

### Taskbar widget
- **Secondary taskbars** — fixed a `StringBuilder.Equals(string)` reference-comparison bug with a proper `OrdinalIgnoreCase` string compare.
- **UIA stalls** — removed blocking `.Wait()` calls (up to 1.5s) on the UI dispatcher during taskbar element lookup; converted to async with timeouts.

### Visualizer
- **Premultiplied alpha** — bar edge pixels are premultiplied before writing to BGRA32, fixing bright fringes on antialiased edges.
- **Bar-count clamp** — deserialized values clamped to 1–20 before resize (prevented a division crash).
- **Stop/restart race** — snapshotting the capture instance instead of double-reading a field being swapped concurrently.
- **Atomic resize** — replacement array is sized and published before `BarCount` changes, avoiding an out-of-bounds window.

### Settings / API / OSD / startup
- **SettingsWindow leak** — the `SettingsManager.PropertyChanged` handler is now unsubscribed on close.
- **Shutdown save** — pending debounced settings save is flushed on exit.
- **Update client** — timeout rotation raised to 3, client swap no longer disposes an in-flight client, User-Agent moved to a per-request header.
- **OSD hide** — candidates are validated against the volume OSD's actual screen position so other shell islands can't be hidden by mistake.
- **Empty monitor list** — flyout animations bail out and schedule a refresh instead of stranding the window off-screen.
- **Shutdown disposal** — audio device monitor and toast activator are properly disposed/unregistered on exit.
- **Run key / lock-key sizing / logs path / purchase dialogs** — null-safety and existence checks to prevent crashes and misleading error states.

## Verification
No .NET SDK available in the authoring environment; verification was review-based (brace balance, no unguarded raw SMTC calls, no blocking UI-thread waits, re-read surrounding logic for each fix). CI (dotnet-format/build) passed on this branch. Manual smoke test before merge still recommended.


Minemetero:
Vibe coded? :<

frans15-2006:
I have read and agree to the FluentFlyout CLA

frans15-2006:
recheck

Issue #870, opened 2026-06-14
[BUG] High CPU Usage


Ritanium:
### Describe the bug

Please fix high usage of CPU in FluentFlyout for no reason. I can't find out why i just use it normal.

### Logs, screenshots & additional context

<img width="678" height="129" alt="Image" src="https://github.com/user-attachments/assets/32746de3-8326-496f-a5cf-d0040429776c" />

### FluentFlyout Version

v2.12.0

### Windows Version

Windows 11 25H2

### Architecture

x64

### Checklist

- [x] I have searched that there is no existing bug report regarding my issue
- [x] I have read the FAQ and there is no answer to my issue

unchihugo:
Hi @Ritanium, thanks for the report, this shouldn't be occuring. Does this spike happen occasionally or all the time you use the program?

eramajarvi:
I can see that spike on my side as well, but only when opening the Settings window of FluentFlyout; it increases the CPU usage to around ~9% then goes down to ~0.2%. Also when navigating in the tabs there is a spike to ~5%
Using W11 26200.8655 with FP 1000.26100.315.0

Ritanium:
Hi @unchihugo, no not all time, it happens when FluentFlyout settings opened in most cases.

Ritanium:
<img width="1284" height="966" alt="Image" src="https://github.com/user-attachments/assets/151d3bd2-ad3e-4a94-9074-e724bb60e829" />

@unchihugo, Settings not opened and still high usage today. it's random

marchingon12:
I can confirm high cpu usage being reported in Task Manager on idle as well, i was wondering why my fans suddenly went 80%. Quitting and restarting the app seems to help.

CPU: Ryzen 7 8845HS
OS: W11 PRO 25H2, 26200.8655
FluentFlyout Version: v2.12.0, bought Premium 
FluentFlyout Settings: https://notebin.de/?6a8cad54b12b9134#2MEdHAgagNzf5dcc6UoaYvchY3tT1LRwcosfSWXUFwQK

unchihugo:
Thanks for the information everyone. Has this started occuring in v2.12.0 or has it always been like this in the past versions?

Ritanium:
It happened for me before v2.12.0

marchingon12:
This has happened even before v.2.12.0.

galagyy:
@unchihugo I saw this a while back and was wondering I could push back the other features for now and take a look at this.

[...]

Issue #827, opened 2026-05-22
[BUG] High CPU usage even when idle


DerDemystifier:
### Describe the bug

It’s at the top of the CPU usage list even when idle, and no music is playing.

The FAQ states:

> There should be negligible or basically 0% CPU usage while running in the background on modern computers. 

That's not the case unfortunately.

I’ve tried disabling the Taskbar widget and other settings, but it had no effect.

### To Reproduce

Cannot reproduce effectively.

### Logs, screenshots & additional context

<img width="933" height="216" alt="Image" src="https://github.com/user-attachments/assets/681196df-ca0a-49f2-84c1-2b85e665950b" />

### FluentFlyout Version

v2.11.2

### Windows Version

Windows 11 Pro 25H2

### Architecture

x64

### Checklist

- [x] I have searched that there is no existing bug report regarding my issue
- [x] I have read the FAQ and there is no answer to my issue

unchihugo:
Hi @DerDemystifier, thanks for reporting. Do you still notice this behavior? If yes, could you list your computer specifications?

DerDemystifier:
```
Processor	AMD Ryzen 5 5600GT
System type	64-bit operating system
```

I’ve figured out how to reproduce the problem—it happens when I press the media buttons. For instance, you can trigger the media track keypress in PowerShell if you don’t have it, and watch Task Manager; it will spike and stay at the top.

```ps
(New-Object -ComObject WScript.Shell).SendKeys([char]176)
```

unchihugo:
Thanks for the information. Does CPU usage still spike if FluentFlyout is closed? If it doesn't, could you try disabling the media flyout to see if the CPU usage is still high then?

DerDemystifier:
Disabling the media flyout doesn’t immediately stop the CPU usage spike. It only drops after I press a media key AFTER disabling the media flyout.

I reckon it has something to do with the media flyout, but it could be some kind of process that persists even after the media flyout slides down.

Is there a logging system in the app? It might be helpful to track metrics alongside events to pinpoint exactly when things went wrong.


I must add that the bug is not reproducible 100% of the time. I only notice it when CoreTemp triggers the overheat protection balloon popup.

lgc2333:
please do not stale this issue

startup
<img width="1242" height="886" alt="Image" src="https://github.com/user-attachments/assets/1d6a96cf-2e37-4c55-9374-db7c01acadc7" />

[...]

PR #1117, opened 2026-09-08
fix: taskbar widget UIA polling causes high CPU (refs #827)

lgc2333:
## Summary

Fix a high-CPU path in the taskbar widget: it walked the taskbar's UI Automation tree on every 1.5s tick and blocked the UI thread with `Wait()`. When the taskbar UIA tree is in a slow state (e.g. during auto-hide taskbar transitions), a single walk takes 1-2.5s, so the app burns almost a full core and the UI stalls.

This is a different trigger than the one in #827. The reporter of #827 has the taskbar widget disabled (their settings export is in the issue thread), and `TaskbarWindow.UpdatePosition()` returns immediately when `TaskbarWidgetEnabled == false` (`TaskbarWindow.xaml.cs:301`), so this code path never ran for them. This PR fixes a path that does burn CPU for widget users; it does not claim to fix the original report.

## Motivation

- Reproduced locally with the widget enabled and the auto-hide taskbar on: 5-8% sustained CPU (~80% of one core), with `Timeout querying taskbar XAML element: WidgetsButton` in the log.
- `dotnet-trace` of the 8% state: `AutomationElement.FindFirst` (~11% inclusive), `ProxyManager.ProxyProviderFromHwnd`, `HwndProxyElementProvider.GetFirstChild`, `Accessible.GetChildren`, `WindowsRebar.GetFirstChild`, plus `DispatcherOperation.WaitOne` (UI thread blocked on the `Wait()` calls).
- Same code path as the one suspected in #870.

References #827

## Type of Change

- [x] Bug fix

## What Changed

- Removed the blocking `findTask.Wait(1000)` / `boundsTask.Wait(500)` calls; the UI thread never waits on automation queries.
- Queries are single-flight and run on a background thread; the cached rect is applied via `Dispatcher.BeginInvoke` when a query completes, so the 1.5s polling picks up the newest element state without blocking.
- Added proportional backoff: if a query takes >300ms, the next one is delayed by 10x its duration (capped at 60s). Without it, background-only queries still keep a core busy in the slow state (measured ~7% total CPU sustained), because each walk itself costs 1-2.5s.
- Element lookup and `BoundingRectangle` are now fetched in one background pass; stale-element handling is preserved (monitor change, disabled element).

## Additional Information

Verified under the same workload that previously sustained 5-8% CPU (auto-hide taskbar cycling + concurrent UIA tree hammering + repeated media keys): CPU spikes only while the flyout is actually displayed and drops back to ~0.1% afterwards. No more `Timeout querying taskbar XAML element` warnings. Widget positioning/normal behavior unchanged.

## Checklist

### DIFF
--- a/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
+++ b/FluentFlyoutWPF/Windows/TaskbarWindow.xaml.cs
@@ -5,6 +5,7 @@ using FluentFlyout.Classes.Settings;
 using FluentFlyoutWPF;
 using FluentFlyoutWPF.Classes;
 using FluentFlyoutWPF.Classes.Utils;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Windows;
@@ -32,16 +33,12 @@ public partial class TaskbarWindow : Window
     private readonly double _scale = 0.9;
 
     private IntPtr _trayHandle;
-    private AutomationElement? _widgetElement;
-    private AutomationElement? _trayElement;
-    private AutomationElement? _taskbarFrameElement;
     // reference to main window for flyout functions
     private MainWindow? _mainWindow;
     private int _lastSelectedMonitor = -1;
     private IntPtr _lastTaskbarHandle;
     private bool _positionUpdateInProgress;
     private bool _isClosing;
-    private readonly Dictionary<string, Task> _pendingAutomationTasks = [];
 
     private GlobalSystemMediaTransportControlsSessionPlaybackStatus? _lastPlaybackStatus;
     private DispatcherTimer? _autoHideTimer;
@@ -360,10 +357,7 @@ on_error:
 
         _lastTaskbarHandle = taskbarHandle;
         _trayHandle = IntPtr.Zero;
-        _widgetElement = null;
-        _trayElement = null;
-        _taskbarFrameElement = null;
-        _pendingAutomationTasks.Clear();
+        _xamlElementStates.Clear();
     }
 
     private void CalculateAndSetPosition(IntPtr taskbarHandle, IntPtr taskbarWindowHandle, bool isMainTaskbarSelected)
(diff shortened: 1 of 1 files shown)


lgc2333:
I have read the CLA Document and I hereby sign the CLA

lgc2333:
dotnet-format passes after the amend. CLAAssistant still fails for a repo-side reason, not the PR:

1. The action cannot write `.github/cla-version1.json` - `master` is covered by the "no merging to main" ruleset (no bypass actors), so it gets `Changes must be made through a pull request` and no signature can ever be recorded (runs 34239760091, 34240430339).
2. The step gate in `.github/workflows/cla-assistant.yml` only accepts the exact comment body `I have read and agree to the FluentFlyout CLA` (or `recheck`), while the bot's message asks for `I have read the CLA Document and I hereby sign the CLA`. My signing comment was therefore skipped (run 34239917264: CLA step skipped).

Fix options: add the OWNER_TOKEN user as a ruleset bypass actor (or store signatures in a separate repo via remote-organization-name/remote-repository-name), and align the gate with the bot's comment text.

lgc2333:
I'm not sure this PR actually fixes the problem raised in the issue, and I apologize if it reads as AI slop. If you don't like it, feel free to close it.

lgc2333:
Because of the above, I'll keep this PR as a draft for now - please review it first. If it makes sense to you, feel free to merge it yourself. Thanks for the time and effort you put into this project.

unchihugo:
Hi @lgc2333, thanks for looking into this. I think instead of a full refactor, we could change the code to just run the UIA method in the background instead of in the UI thread, and let the taskbar polling automatically pick up the newest element states instead of blocking to wait for the elements. I've also fixed the CLA issue, hopefully.

lgc2333:
closing this for now, this maybe an unnecessary fix, sorry

PR #1122, opened 2026-09-09
fix: infinite loop in HideVolumeOsd causes sustained high CPU

### LINKED ISSUE #827: [BUG] High CPU usage even when idle

lgc2333:
> This patch is generated by AI. I'm not familiar with Windows native development, so I can't help that much. I'm really sorry for this. If you mind, close this and make a fix by yourself.

## Summary

`VolumeMixerWindow.HideVolumeOsd()` spins forever when the first enumerated `XamlExplorerHostIslandWindow` is not the volume OSD, pinning one CPU core. This restores the `hwndChildAfter` cursor the port dropped, so the enumeration terminates. Fixes #827.

## Motivation

Closes #827. Trace taken while the reported CPU burn was in progress:

| function | inclusive | exclusive |
| --- | --- | --- |
| `VolumeMixerWindow.HideVolumeOsd()` | 13.87% | 13.87% |
| `VolumeMixerWindow.HideVolumeOsd()` (second capture) | 16.34% | 16.34% |

Equal inclusive and exclusive means it is a leaf: the frame itself is the work, it is not waiting on anything. On a 12-thread machine the reported ~8% is one full core, and nothing else in the process is close.

## Type of Change

- [x] Bug fix

## What Changed

**The cursor.** `FindWindowEx(hwndParent, hwndChildAfter, className, windowName)` — `hwndChildAfter` is a cursor; the search starts at the window *after* that handle. `NULL` means "start at the first child", so the original call returned the same island on every iteration:

```
FindWindowEx(0, 0, "XamlExplorerHostIslandWindow", null), called 6 times:
  call 1..6: 0x304C6   (1 distinct handle out of 6)

FindWindowEx(0, <previous handle>, ...):
  0x304C6 -> 0x8029C -> NULL
```

None of the three `continue` branches advances the cursor, so if the first island fails the rect test the loop has no exit. Passing `hwndXamlIsland` as `hwndChildAfter` walks the islands and terminates. The upstream this file is derived from (`gpkgpk/HideVolumeOSD`, `Windows11` branch) passes the previous `hwndHost`; the port dropped it.

**The removed `if`.** The `if (hwndXamlIsland == IntPtr.Zero) continue;` was inside the loop body, and the loop condition already guarantees the value is non-zero there, so it was unreachable. It reads like a cursor guard but is not one. Cleanup, not part of the fix.

**The all-zero rect guard is unchanged.** It works: the laid-out OSD island's `InputSite` reads `(1163,1378,1163,1378)` while showing and `(1195,1379,1195,1379)` while hidden — degenerate, zero size, but positioned — while the second, never-positioned island reads exactly `(0,0,0,0)` and is skipped. An earlier comment of mine proposed replacing it with `IsWindowVisible`; that was wrong, and the measurements are below.

**Trigger.** `MainWindow.HookCallback` calls `volumeMixerWindow?.ShowFlyout()` on every media/volume key (0xB0-0xB3, 0xAD-0xAF) when `VolumeControlEnabled` is set, which starts `Task.Run(HideVolumeOsd)`. Each further key press stacks another spinning thread.

## Additional Information

**How the CPU was located.** A watcher on the machine samples total CPU every 2s and, when usage stays above 3% for 6 consecutive samples, captures a `dotnet-trace` of the running process. Two captures, both on FluentFlyout 2.15.0 (Store build), both while the burn was still in progress:

| capture | peak CPU | trigger |
| --- | --- | --- |
| 2026-09-09 19:24:17 | 7.75% of total (12 logical cores) | sustained >= 3% for 6s |
| 2026-09-09 19:27:57 | 8.46% of total | sustained >= 3% for 180s |

Both used `Microsoft-DotNETCore-SampleProfiler` (sampled thread time), reported with `dotnet-trace report ... topN`. `dotnet-stack` on the live process gave two managed threads marked `CPU_TIME`, the same shape in both captures:

```
Thread (0x9AC):            Thread (0x540C):
  CPU_TIME                   CPU_TIME
  FluentFlyout!FluentFlyoutWPF.Windows.VolumeMixerWindow.HideVolumeOsd()
  FluentFlyout!FluentFlyoutWPF.Windows.VolumeMixerWindow+<>c.<ShowFlyout>b__16_0()
  System.Threading.Tasks.Task.InnerInvoke()
  System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(...)
  System.Threading.Tasks.Task.ExecuteWithThreadLocal(...)
  System.Threading.ThreadPoolWorkQueue.Dispatch()
```

**Upstream comparison.** The source this follows is the `Windows11` branch of `gpkgpk/HideVolumeOSD` (it defines the same three window classes and carries the same zero-rect comment), and the differences are adaptations, not inventions:

| | upstream (`Windows11`) | this port |
| --- | --- | --- |
| rect measured | `GetClientRect(hwndHost)` — the island | `GetWindowRect(hwndInputClass)` — the `InputSite` child |
| restored before measuring | `ShowWindow(hwndHost, 9)` | `ShowWindow(hwndInputClass, 9)` |
| multiple matches | counts, errors on 0 or 2+ | takes the first |

On a current build the port's variant discriminates better: upstream's host-client-rect test returns non-zero for *both* islands here (`(0,0,169,61)` for the OSD, `(0,0,2560,1440)` for the full-screen one), so it would hit its own "Multiple pairs found!" error, while the port's `InputSite` test excludes the second cleanly. That is why the guard is left alone and only the cursor is restored.

**Corrections to my earlier comments.** Two claims I made were wrong and are retracted here: that the rect guard cannot discriminate (it can — table above), and that `IsWindowVisible` should replace it (it cannot discriminate — two runs with FluentFlyout stopped gave `true/true` in one and `false/false` in the other for the same two islands).

**One loose end, for the maintainer.** `ShowWindow(hwndInputClass, SW_RESTORE)` does not restore what this app minimizes: the success path ends with `ShowWindow(_nativeOsdElement, SW_MINIMIZE)`, i.e. the host island. Measured after a FluentFlyout-style hide:

| step | host iconic | host clientRect | `InputSite` winRect |
| --- | --- | --- | --- |
| after `ShowWindow(host, SW_MINIMIZE)` | true | `(0,0,0,0)` | `(0,1412,0,1412)` |
| after `ShowWindow(input, SW_RESTORE)` (this port) | true | `(0,0,0,0)` | `(0,1412,0,1412)` |
| after `ShowWindow(host, SW_RESTORE)` (upstream) | false | `(0,0,169,61)` | `(-32768,-32768,-32768,-32768)` |

The `InputSite` rect is non-zero either way, and applying `SW_RESTORE` to the `InputSite` of a never-positioned island does not make its rect non-zero, so the call looks inert for this predicate. Upstream needs its host restore because it reads the host's client rect, which minimize zeroes out. If that line is doing something I am not seeing, I would like to know. (The `if (GetWindowRect(...))` also has no `else`, so a failure neither breaks nor clears `hwndOsd` — inherited from upstream, never observed to fail.)

**Where that leaves this.** That locates the main cause with reasonable confidence; what to do about the loop is the maintainer's call. If this patch is not how you want it, fix it your own way and close this PR — I have no attachment to it.

**Disclosure.** I found the root cause and produced the traces; the patch itself was written with AI assistance. I reproduced the loop and ran the fixed end-to-end build myself, but I cannot answer questions about Windows internals beyond what is written here.

## Checklist
- [x] AI tools were used (I reviewed and fully understand the changes myself).

### DIFF
--- a/FluentFlyoutWPF/Windows/VolumeMixerWindow.xaml.cs
+++ b/FluentFlyoutWPF/Windows/VolumeMixerWindow.xaml.cs
@@ -197,15 +197,11 @@ public partial class VolumeMixerWindow : MicaWindow
     // derived from gpkgpk/HideVolumeOSD: https://github.com/gpkgpk/HideVolumeOSD
     private static void HideVolumeOsd()
     {
-        // find widget in XAML
-        IntPtr hwndXamlIsland, hwndOsd = IntPtr.Zero;
-        while ((hwndXamlIsland = FindWindowEx(IntPtr.Zero, IntPtr.Zero, "XamlExplorerHostIslandWindow", null)) != IntPtr.Zero)
+        // find widget in XAML; FindWindowEx must be given the previous handle as hwndChildAfter,
+        // otherwise it returns the same first island forever and this loop never terminates
+        IntPtr hwndXamlIsland = IntPtr.Zero, hwndOsd = IntPtr.Zero;
+        while ((hwndXamlIsland = FindWindowEx(IntPtr.Zero, hwndXamlIsland, "XamlExplorerHostIslandWindow", null)) != IntPtr.Zero)
         {
-            if (hwndXamlIsland == IntPtr.Zero)
-            {
-                continue;
-            }
-
             hwndOsd = FindWindowEx(hwndXamlIsland, IntPtr.Zero, "Windows.UI.Composition.DesktopWindowContentBridge", "DesktopWindowXamlSource");
             if (hwndOsd == IntPtr.Zero)
             {
@@ -220,7 +216,7 @@ public partial class VolumeMixerWindow : MicaWindow
                 continue;
             }
 
-            ShowWindow(hwndInputClass, 9); // SW_RESTORE
+            ShowWindow(hwndInputClass, SW_RESTORE);
             if (GetWindowRect(hwndInputClass, out RECT rect))
             {
                 if (rect.Top == 0 && rect.Left == 0 && rect.Bottom == 0 && rect.Right == 0)


lgc2333:
> This comment is generated by AI. I reviewed the code and intuitively felt that this fix might be too straightforward, so I continued investigating. Because i'm not familiar with Windows native development, I'm not sure what he is saying. Sorry for that.

Following up on my own patch. **Corrected** — an earlier version of this comment claimed the zero-rect guard does not discriminate and proposed replacing it with `IsWindowVisible`. That was wrong. I re-measured with FluentFlyout stopped and the guard does work. The patch itself (restore the `hwndChildAfter` cursor) stands.

**The guard.** Islands on this machine, FluentFlyout stopped, all owned by explorer:

| island | InputSite `GetWindowRect` | guard verdict |
| --- | --- | --- |
| OSD island, native OSD shown | `(1163,1378,1163,1378)` | non-zero → selected |
| OSD island, OSD hidden | `(1195,1379,1195,1379)` | non-zero → selected |
| second island (never positioned) | `(0,0,0,0)` | all-zero → skipped |

The OSD's own rect is degenerate — zero width and zero height, it is a hit-test region — but it is *positioned*, and that is what the guard keys on. So it does separate the laid-out OSD island from never-positioned islands. My note that a `width > 0` variant would be wrong still holds; my conclusion that the guard is useless does not.

**Why not `IsWindowVisible`.** Measured with FluentFlyout stopped, two runs: both islands read `visible=True` in one run, both read `visible=False` in the other. It does not separate the islands, and switching to it would make the code fail to find the OSD. Suggestion withdrawn.

**Upstream, corrected.** The source this port follows is the **`Windows11` branch** of `gpkgpk/HideVolumeOSD`, not the default branch — that branch defines exactly the class names used here (`XamlExplorerHostIslandWindow`, `Windows.UI.Composition.DesktopWindowContentBridge`, `Windows.UI.Input.InputSite.WindowClass`) and carries the same zero-rect comment. So the intent is inferable, and the differences are adaptations rather than inventions:

| | upstream (`Windows11`) | this port |
| --- | --- | --- |
| rect measured | `GetClientRect(hwndHost)` — the island | `GetWindowRect(hwndInputClass)` — the InputSite child |
| restored before measuring | `ShowWindow(hwndHost, 9)` | `ShowWindow(hwndInputClass, 9)` |
| multiple matches | counts, errors on 0 or 2+ | takes the first |

On a current build the port's variant discriminates *better*: upstream's host-client-rect test returns non-zero for both islands here (`(0,0,169,61)` for the OSD, `(0,0,2560,1440)` for the full-screen one), so it would hit its own "Multiple pairs found!" error, while the port's InputSite test excludes the second cleanly. That looks like a deliberate adaptation to builds that have more than one island. Which is also why the dropped cursor — an obvious defect on its own — is the only thing I would change here.

**One loose end, for you.** `ShowWindow(hwndInputClass, SW_RESTORE)` does not restore what this app minimizes: the success path ends with `ShowWindow(_nativeOsdElement, SW_MINIMIZE)`, i.e. the *host island*. Measured after a FluentFlyout-style hide:

| step | host iconic | host clientRect | InputSite winRect |
| --- | --- | --- | --- |
| after `ShowWindow(host, SW_MINIMIZE)` | true | `(0,0,0,0)` | `(0,1412,0,1412)` |
| after `ShowWindow(input, SW_RESTORE)` (this port) | true | `(0,0,0,0)` | `(0,1412,0,1412)` |
| after `ShowWindow(host, SW_RESTORE)` (upstream) | false | `(0,0,169,61)` | `(-32768,-32768,-32768,-32768)` |

The InputSite rect stays non-zero either way, and applying `SW_RESTORE` to the InputSite of a never-positioned island does not make its rect non-zero. So the call looks inert for this predicate. Upstream needs its host restore because upstream reads the host's client rect, which minimize does zero out. Is that line doing something I am not seeing?

(The `if (GetWindowRect(...))` still has no `else`, so a failure neither breaks nor clears `hwndOsd`. It never failed in any run here — inherited from upstream, theoretical.)

lgc2333:
hi, just a friendly reminder because there's no response for a while. review this if you have time :)

unchihugo (commented):
Hi @lgc2333, what does adding a second parameter in `FindWindowEx(IntPtr.Zero, hwndXamlIsland, "XamlExplorerHostIslandWindow", null))` do, and why was the if statement removed? I'd like to know the thought process!

lgc2333:
hi, i have only digged out the root cause of the cpu usage issue, but the fix is provided by AI. so i can't answer this. sorry for that.
but the only thing i know is the code there is different from upstream, why did you make these changes?
close this and make your own fix, if you mind. sorry again.

lgc2333:
@unchihugo thanks for looking at this. Answering both questions, then correcting something I got wrong.

**What the second parameter does.** `FindWindowEx(hwndParent, hwndChildAfter, className, windowName)` — `hwndChildAfter` is a cursor: the search starts at the window *after* that handle in Z order. `NULL` means "start at the first child", so `FindWindowEx(Zero, NULL, ...)` returns the same first island on every call.

That is the entire bug. The loop body re-runs on one window forever, because none of the three `continue` branches moves the cursor forward:

- `hwndOsd == Zero` → `continue`, cursor unchanged
- `hwndInputClass == Zero` → `continue`, cursor unchanged
- rect is all-zero → `continue`, cursor unchanged

If the first island is not the volume OSD, no branch ever makes progress. Passing the handle from the previous iteration turns the lookup into a walk over the islands, and it returns `NULL` when they run out. This is the standard enumeration pattern, and it is what the code this file is derived from does: in `gpkgpk/HideVolumeOSD` (the `Windows11` branch), the equivalent loop passes the previous `hwndHost` as the second argument. The port dropped it.

**Why the `if` was removed.**

```csharp
while ((hwndXamlIsland = FindWindowEx(...)) != IntPtr.Zero)
{
    if (hwndXamlIsland == IntPtr.Zero) continue;
```

The loop condition already guarantees the value is non-zero when the body runs, so that `continue` is unreachable. It reads like a cursor guard but is not one. Removing it was cleanup, not part of the fix.

**Correction — my earlier worry about the rect guard was wrong.** In an earlier comment I claimed the `rect == (0,0,0,0)` test cannot tell the OSD island from any other, and proposed `IsWindowVisible` instead. I have since re-measured with FluentFlyout stopped, and both claims fail:

- The guard does discriminate. The laid-out OSD island's InputSite reads non-zero (`(1163,1378,1163,1378)` while the OSD is showing, `(1195,1379,1195,1379)` while it is hidden — degenerate, zero size, but positioned), while the second, never-positioned island reads exactly `(0,0,0,0)` and is skipped.
- `IsWindowVisible` does *not* discriminate. Two runs with FluentFlyout stopped gave `true/true` in one and `false/false` in the other, for the same two islands. Using it would break the lookup.

I have rewritten the longer comment above with the corrected measurements and the upstream comparison. Net result: the only change I would make to this loop is restoring the `hwndChildAfter` cursor. The guard can stay as it is.

The one loose end left is `ShowWindow(hwndInputClass, SW_RESTORE)`. It restores the InputSite child, but the window this app minimizes is the host island — the success path ends with `ShowWindow(_nativeOsdElement, SW_MINIMIZE)`. The InputSite rect reads non-zero either way, so the call looks inert for this predicate. If you know why it targets the child rather than the host, I would like to hear it.

lgc2333:
Adding the raw evidence for the claim in the PR description, since that is the part worth checking.

**One line: the CPU went into `VolumeMixerWindow.HideVolumeOsd()`.**

**How that was located.** I ran a small watcher on the machine that samples total CPU every 2s and, when usage stays above 3% for 6 consecutive samples, grabs a `dotnet-trace` of the running process. Two captures, both on FluentFlyout 2.15.0 (Store build), both while the CPU burn was still in progress:

| capture | peak CPU | reason |
| --- | --- | --- |
| 2026-09-09 19:24:17 | 7.75% of total (12 logical cores) | sustained >= 3% for 6s |
| 2026-09-09 19:27:57 | 8.46% of total | sustained >= 3% for 180s |

On a 12-thread machine 8% is one full core. Both traces were taken with `Microsoft-DotNETCore-SampleProfiler` (sampled thread time), analysed with `dotnet-trace report ... topN`:

| function | inclusive | exclusive |
| --- | --- | --- |
| `VolumeMixerWindow.HideVolumeOsd()` | 13.87% | 13.87% |
| `VolumeMixerWindow.HideVolumeOsd()` (second capture) | 16.34% | 16.34% |

Equal inclusive and exclusive means it is a leaf — the frame itself is the work; it is not waiting on anything. Nothing else in the process is close.

**Stack, both captures.** `dotnet-stack` on the live process gives two managed threads marked `CPU_TIME`, same shape both times:

```
Thread (0x9AC):            Thread (0x540C):
  CPU_TIME                   CPU_TIME
  FluentFlyout!FluentFlyoutWPF.Windows.VolumeMixerWindow.HideVolumeOsd()
  FluentFlyout!FluentFlyoutWPF.Windows.VolumeMixerWindow+<>c.<ShowFlyout>b__16_0()
  System.Threading.Tasks.Task.InnerInvoke()
  System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(...)
  System.Threading.Tasks.Task.ExecuteWithThreadLocal(...)
  System.Threading.ThreadPoolWorkQueue.Dispatch()
```

So it is thread-pool work started by the `ShowFlyout` lambda, spinning, and each key press adds another copy.

**Why it spins.** Deterministic, on the running desktop, no debugger needed:

```
FindWindowEx(0, 0, "XamlExplorerHostIslandWindow", null)  called 6 times:
  call 1: 0x304C6
  call 2: 0x304C6
  call 3: 0x304C6
  call 4: 0x304C6
  call 5: 0x304C6
  call 6: 0x304C6
  --> distinct handles returned: 1 of 6

FindWindowEx(0, <previous handle>, ...) :
  0x304C6  InputSite rect=(1195,1379,1195,1379)
  0x8029C  InputSite rect=(0,0,0,0)
  --> 2 islands, then NULL
```

With `hwndChildAfter` left at `NULL`, the loop re-reads the same handle no matter which branch it took, so it has no way out when the first island fails the rect test. With the cursor carried forward it walks the islands and ends. That is the whole defect, and it reproduces on demand.

**Where that leaves this.** That locates the main cause with reasonable confidence; what to do about the loop is your call. If the patch here is not how you want it, feel free to fix it your own way and close this PR — I have no attachment to it.

Sorry for the AI slop on this one. The bug and the evidence are real, but I should have brought the evidence first and the patch second.

lgc2333:
All of the minimized comments above have been folded into the PR description — nothing in them was dropped. Sorry for the wall of long comments, and sorry for the email spamming.

unchihugo (approved):
Sounds good, thanks.

lgc2333:
Bump. Still reproducing on my machine: with the loop present the native save dialog takes 2.4-2.9 s to appear, and with FluentFlyout stopped it drops to 0.31-0.55 s. Would like to see this merged and released soon.

